PR #6, opened 2020-10-06
I moved a lot around :O

m6jones:


framework/lit/HypermediaStateMixin.js
```diff
@@ -6,7 +6,7 @@ export { observableTypes } from '../state/HypermediaState.js';
  * @export
  * @polymerMixin
  **/
-export const HypermediaLitMixin = superclass => class extends superclass {
+export const HypermediaStateMixin = superclass => class extends superclass {
```
m6jones, line 9: I changed the name since why do we need lit in the title when it is in the lit folder :D

ciss1995:
LGTM, you just might need to change the path in the test file for the build to be green

ciss1995:
🎉

ciss1995 (approved):

PR #9, opened 2020-10-13
testing state

AKobets:
Adding some test for StateStore

### DIFF
--- a/test/state/store.test.js
+++ b/test/state/store.test.js
@@ -26,6 +26,8 @@ describe('state/store.js', () => {
 		expect(state).to.be.instanceOf(HypermediaState);
 		expect(state.entityId, 'entityId').to.be.equal(id);
 		expect(state.token.value, 'token').to.be.equal(token);
+		const storedState = window.D2L.SirenSdk.StateStore.get(id, token);
+		expect(storedState, 'stored state').to.be.equal(state); // BUG, returns false instead of state object for state that is created with TOKEN_COOKIE
 	}
 
 	beforeEach(() => {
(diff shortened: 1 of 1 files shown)


test/state/store.test.js
```diff
@@ -0,0 +1,212 @@
...
+		sandbox = sinon.createSandbox();
+	});
+
+	afterEach(() => {
+		emptyRecycleBin();
+		sandbox.restore();
+	});
+
+	describe('stateFactory', () => {
+		/*
```
AKobets, line 43: will remove all Testing notes comments that not needed. 
For now I use them to keep track of my questions while learning code and creating tests

test/state/store.test.js
```diff
@@ -0,0 +1,212 @@
...
+			it(`should retun undefined when created with id:  ${JSON.stringify(test.id)}`, async() => {
+				const state = await createTestState(undefined, 'token');
+				expect(state).to.be.equal(undefined);
+			});
+		});
+
+		describe('integration with token', () => {
+			it('can create state with TOKEN_COOKIE', async() => {
+				const state = await createTestState('id', TOKEN_COOKIE);
+				verifyState(state, 'id', TOKEN_COOKIE);
```
AKobets, line 89: state could be created with TOKEN_COOKIEbut could not be r etrieved later with SateStore class method get()

zinabat: Excellent - new tests are finding bugs!

AKobets: same problem when cerating state with undefined token, the state is created but get throws. See failing tests in git actions logs
```
 ✖ can create state with TOKEN_COOKIE
        ✔ can create state with function token
        ✔ can create state with function that returns promise
        ✖ can create new state with undefined token
```

AKobets:
there is another PR in draft state that tests most recent implementation.

PR #11, opened 2020-10-20
Ccisse/state factory

ciss1995:


state/stateFactory.js
```diff
@@ -0,0 +1,228 @@
...
+import { getEntityIdFromSirenEntity } from './sirenComponents/Common.js';
+import SirenParse from 'siren-parser';
+
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.fetchStatus = window.D2L.SirenSdk.fetchStatus || new Map();
+window.D2L.SirenSdk.d2lfetch = window.D2L.SirenSdk.d2lfetch || window.d2lfetch;
+window.D2L.SirenSdk.StateStore = window.D2L.SirenSdk.StateStore || new StateStore();
+
+export async function stateFactory(options) {
```
zinabat, line 13: You can use ES6 destructuring to do this more easily, e.g,
```js
export async function stateFactory({state, token, rawEntity, entityID}) {
    if (state) //etc
```
However, I'd like to see if we can make the usage for this more obvious/readable. There must be a way to have the function auto-detect when we are passing a half-fetched state vs an href and token vs a ready-made state and do what it needs to do without telling it explicitly.

zinabat (commented):
First pass, I'll look at this more later

state/sirenComponents/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
+export class ObserverMap {
```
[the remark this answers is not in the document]
ciss1995: Oh yes good catch,  this is for another story.

state/stateFactory.js
ciss1995: We could just have 2 vars one for token and one for the rest and check constructor.name and switch accordingly?

m6jones: Wait, why do we need state? This makes states.

framework/lit/HypermediaResult.js
```diff
@@ -199,10 +199,10 @@ export class HypermediaResult extends TemplateResult {
 		const resources = { classes: [] };
 		const components = componentStoreFactory(pseudoTag);
 		if (!href || !token) return html`loading`;
-		const statePromise = stateFactory(href, token);
+		const statePromise = stateFactory(href.constructor.name, href, token);
 		const fetchedResults  = statePromise.then(async state => {
 			state.addObservables(resources, observable);
-			await fetch(state);
+			await stateFactory(state.constructor.name, state);
```
m6jones, line 205: This should still be a fetch call. We shouldn't need to call the stateFactory twice.

framework/lit/HypermediaStateMixin.js
```diff
@@ -54,9 +54,9 @@ export const HypermediaStateMixin = superclass => class extends superclass {
 		if (this.__gettingState) return;
 		try {
 			this.__gettingState = true;
-			this._state = await stateFactory(this.href, this.token);
+			this._state = await stateFactory(this.href.constructor.name, this.href, this.token);
 			this._state.addObservables(this, this._observables);
-			await fetch(this._state);
+			await stateFactory(this._state.constructor.name, this._state);
```
m6jones, line 59: Same this is still fetch.

state/sirenComponents/SirenLink.js
```diff
@@ -63,11 +63,11 @@ export class SirenLink {
 		}
 
 		if (this._token) {
-			this._childState = await stateFactory(this.link.href, shouldAttachToken(this._token, this.link));
+			this._childState = await stateFactory(this.link.href.constructor.name, this.link.href, shouldAttachToken(this._token, this.link));
 			this._routes.forEach((route, component) => {
 				this._childState.addObservables(component, route);
 			});
-			fetch(this._childState);
+			stateFactory(this._childState.constructor.name, this._childState);
```
m6jones, line 70: same this should remain fetch.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+import { getEntityIdFromSirenEntity } from './sirenComponents/Common.js';
+import { HypermediaState } from './HypermediaState.js';
+import { StateStore } from './store.js';
+import SirenParse from 'siren-parser';
+
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.fetchStatus = window.D2L.SirenSdk.fetchStatus || new Map();
+window.D2L.SirenSdk.d2lfetch = window.D2L.SirenSdk.d2lfetch || window.d2lfetch;
+window.D2L.SirenSdk.StateStore = window.D2L.SirenSdk.StateStore || new StateStore();
```
m6jones, line 12: Lets remove SirenSdk from here.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+import { getToken, shouldAttachToken } from './token.js';
+import { getEntityIdFromSirenEntity } from './sirenComponents/Common.js';
+import { HypermediaState } from './HypermediaState.js';
+import { StateStore } from './store.js';
+import SirenParse from 'siren-parser';
+
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.fetchStatus = window.D2L.SirenSdk.fetchStatus || new Map();
+window.D2L.SirenSdk.d2lfetch = window.D2L.SirenSdk.d2lfetch || window.d2lfetch;
```
m6jones, line 11: I don't think we need this.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
+import 'd2l-fetch/d2l-fetch.js';
+import { getToken, shouldAttachToken } from './token.js';
+import { getEntityIdFromSirenEntity } from './sirenComponents/Common.js';
+import { HypermediaState } from './HypermediaState.js';
+import { StateStore } from './store.js';
+import SirenParse from 'siren-parser';
+
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.fetchStatus = window.D2L.SirenSdk.fetchStatus || new Map();
```
m6jones, line 10: We don't need this either. fetchStatus will be held within fetchable classes.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+window.D2L = window.D2L || {};
+window.D2L.SirenSdk = window.D2L.SirenSdk || {};
+window.D2L.SirenSdk.fetchStatus = window.D2L.SirenSdk.fetchStatus || new Map();
+window.D2L.SirenSdk.d2lfetch = window.D2L.SirenSdk.d2lfetch || window.d2lfetch;
+window.D2L.SirenSdk.StateStore = window.D2L.SirenSdk.StateStore || new StateStore();
+
+export async function stateFactory(type, objParam, token) {
+	switch (type) {
+		case 'HypermediaState':
+			return _fetch(objParam);
```
m6jones, line 17: stateFactory shouldn't fetch.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+		case 'Link':
+			return _stateFactoryByRawSirenEntity(objParam, token);
+		case 'String':
+			return _stateFactory(objParam, token);
```
m6jones, line 21: HyepermediaState constructor can worry about if it was feed some data and how to get the entityId. We shouldn't need to tell it about what type.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+		await _handleCachePriming(action.token, response);
+		const json = await response.json();
+		const entity = await SirenParse(json);
+		const state = window.D2L.SirenSdk.fetchStatus.get(entity.getLink('self').href, action.token);
+
+		state.onServerResponse(entity);
+	} catch (err) {
+		console.error(err);
+	}
+}
```
m6jones, line 52: This can be removed.

state/stateFactory.js
```diff
@@ -0,0 +1,215 @@
...
+
+export async function refreshState(state, refetch = true) {
+	await state.refreshToken();
+	return _fetch(state, refetch);
+}
```
m6jones, line 57: Remove

m6jones (commented):
It might be a better idea to go for the golden product but not connect it yet. I can connect it with the fetch.

So basically, the store needs to hold `HypermediaState` objects keyed by `token` (`token.toString`) and `href`

My screw up when I wrote the story :( Otherwise this was awesome!

state/stateFactory.js
m6jones: My mind is arguing with this point. Maybe this is the place to parse out the entityId from an entity.

ciss1995:
I believe this reflects what we want better https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/13

PR #13, opened 2020-10-22
Create a container that holds hypermediaState obj

ciss1995:


state/store.js
```diff
@@ -6,6 +6,7 @@ import SirenParse from 'siren-parser';
 
 class StateStore {
 	constructor(fetch) {
+		this.states = new Map();
```
tenbradley, line 9: Is the intent to have both this.states and this._states?

state/store.js
```diff
@@ -264,3 +269,13 @@ export async function performAction(action, input) {
 export async function dispose(state, component) {
 	state && state.dispose(component);
 }
+
+function _addStateToMap(href, token, state) {
```
m6jones, line 273: This is probably more personal preference. 

I would rather the `token` object get passed here and we use the `token.toString()` in this method.

When we come back to it, that will make all the `token` variables represent the same thing.

state/store.js
ciss1995: Humm good question. Both are doing a similar thing but slightly different. The `_states` stores the state as a key and count as val and state stores the state as value. We could definitely just combine the two if that's the direction we want to take

state/store.js
ciss1995: Yeah, definitely something we can do

state/store.js
tenbradley: We can have both. I would just suggest less similar names.

ciss1995: Good point, I'll rename the new one.

state/store.js
```diff
@@ -6,6 +6,7 @@ import SirenParse from 'siren-parser';
 
 class StateStore {
 	constructor(fetch) {
+		this.cachedStates = new Map();
```
m6jones, line 9: This should be private?

state/store.js
```diff
@@ -221,23 +222,27 @@ export async function refreshState(state, refetch = true) {
 	return window.D2L.SirenSdk.StateStore.fetch(state, refetch);
 }
 
-export function stateFactoryByRawSirenEntity(rawEntity, token) {
-	const entityId = getEntityIdFromSirenEntity(rawEntity);
+export async function stateFactory(entity, token) {
+	if (!entity) return;
+
+	const entityId = entity.constructor.name === 'Link' ? getEntityIdFromSirenEntity(entity) : entity;
```
m6jones, line 228: I would do this this other way. Where I'd check to see if entity is a string. If it isn't run it through `getEntityIdFromSirenEntity`

m6jones:
I updated the notes in the story:
https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F441183098312

I am going to split this into two stories. This is my bad :( 

Discussed with Ten and Lloyd on the call.

ciss1995:
This should be replaced by https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/19 and https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/18.

PR #15, opened 2020-10-26
Fetchable mixin

zinabat:
## Context
[US121368](https://rally1.rallydev.com/#/detail/userstory/441183090652?fdp=true): Creation of Fetchable mixin
[Architecture Design](https://desire2learn.atlassian.net/wiki/spaces/POL/pages/1622278478/US120980+Defining+the+architecture+of+hmc-foundation)
Adds the `Fetchable` mixin which functions as an abstract class for use with the in progress `fetch` method.

- `fetchStatus` is now an object attached to the `Fetchable` with `start()`, `cancel()`, and `done()` methods
- `observable/SirenAction` and `HypermediaState` both implement `Fetchable`
- `href(input)` method converted to `setQueryParams` method
- `header()` method converted to `headers` getter
- Snuck a `.editorconfig` file in there because it was causing problems not having one

## Quality
- added new unit tests for `FetchStatus`
- added basic tests for `Fetchable.setQueryParams()`
- manually tested with BSI branch

**Note**
There is an existing error in `master` related to `sirenAction`. I didn't want to fix it in this PR because of the current work in flight.
Screen Shot 2020-10-26 at 2 58 46 PM

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf8
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab
+indent_size = 4
+
+[{*.json,*.yml}]
+indent_style = space
+indent_size = 2
--- a/state/observable/SirenAction.js
+++ b/state/observable/SirenAction.js
@@ -32,6 +32,7 @@ export class SirenAction extends Fetchable(Observable) {
 	}
 
 	get headers() {
+		super.headers;
 		// set header content to json if present
 		if (this._rawSirenAction.type.indexOf('json') !== -1) {
 			this._headers.set('Content-Type', this._rawSirenAction.type);
--- a/state/observable/SirenAction.js
+++ b/state/observable/SirenAction.js
@@ -4,7 +4,7 @@ import { performAction } from '../store.js';
 
 export class SirenAction extends Fetchable(Observable) {
 	constructor({ id: name, token, state }) {
-		super(state?.href, token);
+		super(null, token);
 		this._action = { has: false, perform: () => undefined, update: () => undefined };
 		this._name = name;
 		this._state = state;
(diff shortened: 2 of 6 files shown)


state/observable/SirenAction.js
```diff
@@ -1,13 +1,12 @@
+import { Fetchable } from '../Fetchable.js';
 import { Observable } from './Observable.js';
 import { performAction } from '../store.js';
-import { refreshToken } from '../token.js';
 
-export class SirenAction extends Observable {
+export class SirenAction extends Fetchable(Observable) {
 	constructor({ id: name, token, state }) {
-		super();
+		super(state?.href, token);
```
zinabat, line 7: I think the state should always exist? I wasn't sure.

state/observable/SirenAction.js
```diff
@@ -25,51 +24,43 @@ export class SirenAction extends Observable {
...
-
-		return headers;
+	get method() {
+		return this._rawSirenAction && this._rawSirenAction.method;
 	}
 
-	href(input) {
-		let url = new URL(this._rawSirenAction.href, window.location.origin);
-		if (this._rawSirenAction.method === 'GET' || this._rawSirenAction.method === 'HEAD') {
+	setBodyFromInput(input) {
```
zinabat, line 47: I'd like to add tests for this, but I'm unsure what the input is supposed to look like.

state/observable/SirenAction.js
m6jones: defineProperties should make it always exist.

m6jones: This shouldn't be the `href` that we should put to super.

zinabat: Should the `href` be null then until the `setSirenEntity` gets us the `_rawSirenAction`?

tenbradley: That logically makes sense to me.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
```
m6jones, line 1: Thank you!

state/observable/SirenAction.js
m6jones: yup

state/observable/SirenAction.js
m6jones: Lets figure out tests when I setup fetch method.

tenbradley:
Should we update the variables that are "basicInfo" to reflect the name change to "definedProperty"?

tenbradley (approved):

PR #16, opened 2020-10-27
Rename component to ObserverMap

ciss1995:
Context: US121366
- Rename Component to ObserverMap
- Add value property to hold the value that will be passed in
- Add test

### DIFF
--- a/state/observable/Observable.js
+++ b/state/observable/Observable.js
@@ -1,16 +1,14 @@
-import { Component } from './Common.js';
+import { ObserverMap } from './ObserverMapjs';
 //import { defaultBasicInfo } from 'sirenComponentFactory.js';
 
 export class Observable {
 
-	// TODO: change to ObserverMap in US121366
 	constructor() {
-		this._observers = new Component(); // new ObserverMap()
+		this._observers = new ObserverMap();
 	}
 
 	addObserver(observer, property, { method }, value) {
-		this._observers.add(observer, property, method);
-		this._observers.setObserverProperty(observer, value);
+		this._observers.add(observer, property, method, value);
 	}
 
 	deleteObserver(observer) {
--- a/state/observable/Observable.js
+++ b/state/observable/Observable.js
@@ -1,4 +1,4 @@
-import { ObserverMap } from './ObserverMapjs';
+import { ObserverMap } from './ObserverMap.js';
 
 export class Observable {
 
--- a/state/observable/Observable.js
+++ b/state/observable/Observable.js
@@ -6,8 +6,8 @@ export class Observable {
 		this._observers = new ObserverMap();
 	}
 
-	addObserver(observer, property, { method }, value) {
-		this._observers.add(observer, property, method, value);
+	addObserver(observer, property, { method }) {
+		this._observers.add(observer, property, method);
 	}
 
 	deleteObserver(observer) {
--- a/state/observable/Observable.js
+++ b/state/observable/Observable.js
@@ -1,5 +1,4 @@
 import { ObserverMap } from './ObserverMapjs';
-//import { defaultBasicInfo } from 'sirenComponentFactory.js';
 
 export class Observable {
 
--- a/state/observable/ObserverMap.js
+++ b/state/observable/ObserverMap.js
@@ -2,10 +2,10 @@ export class ObserverMap {
 	constructor() {
 		this._observers = new Map();
 		this._methods = new WeakMap();
-		this.value = {};
+		this.value = undefined;
 	}
 
-	add(observer, property, method, value) {
+	add(observer, property, method) {
 		if (this._observers.has(observer)) {
 			return;
 		}
@@ -13,7 +13,6 @@ export class ObserverMap {
 		this.value = value;
 		this._observers.set(observer, property);
 		method && this._methods.set(observer, method);
-		this._setObserverProperty(observer, value);
 	}
 
 	delete(observer) {
--- a/state/observable/ObserverMap.js
+++ b/state/observable/ObserverMap.js
@@ -5,34 +5,34 @@ export class ObserverMap {
 		this.value = {};
 	}
 
-	add(component, property, method, value) {
-		if (this._observers.has(component)) {
+	add(observer, property, method, value) {
+		if (this._observers.has(observer)) {
 			return;
 		}
 
 		this.value = value;
-		this._observers.set(component, property);
-		method && this._methods.set(component, method);
-		this._setObserverProperty(component, value);
+		this._observers.set(observer, property);
+		method && this._methods.set(observer, method);
+		this._setObserverProperty(observer, value);
 	}
 
-	delete(component) {
-		this._methods.delete(component);
-		return this._observers.delete(component);
+	delete(observer) {
+		this._methods.delete(observer);
+		return this._observers.delete(observer);
 	}
 
 	setProperty(value) {
 		this.value = value;
 
-		this._observers.forEach((property, component) => {
-			const method = this._methods.has(component) && this._methods.get(component);
-			component[property] = method ? method(this.value) : this.value;
+		this._observers.forEach((property, observer) => {
+			const method = this._methods.has(observer) && this._methods.get(observer);
+			observer[property] = method ? method(this.value) : this.value;
 		});
 	}
 
-	_setObserverProperty(component) {
-		const method = this._methods.has(component) && this._methods.get(component);
-		component[this._observers.get(component)] = method ? method(this.value) : this.value;
+	_setObserverProperty(observer) {
+		const method = this._methods.has(observer) && this._methods.get(observer);
+		observer[this._observers.get(observer)] = method ? method(this.value) : this.value;
 	}
 
 }
--- a/state/observable/ObserverMap.js
+++ b/state/observable/ObserverMap.js
@@ -1,0 +1,43 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = {};
+	}
+
+	add(component, property, method, value) {
+		if (this._observers.has(component)) {
+			return;
+		}
+
+		this.value = value;
+		this._observers.set(component, property);
+		method && this._methods.set(component, method);
+		this._setObserverProperty(component, value);
+	}
+
+	delete(component) {
+		this._methods.delete(component);
+		return this._observers.delete(component);
+	}
+
+	setProperty(value) {
+		this.value = value;
+
+		this._observers.forEach((property, component) => {
+			const method = this._methods.has(component) && this._methods.get(component);
+			component[property] = method ? method(this.value) : this.value;
+		});
+	}
+
+	_setObserverProperty(component) {
+		const method = this._methods.has(component) && this._methods.get(component);
+		component[this._observers.get(component)] = method ? method(this.value) : this.value;
+	}
+
+}
+
+export function getEntityIdFromSirenEntity(entity) {
+	const self = entity.hasLinkByRel && entity.hasLinkByRel('self') && entity.getLinkByRel && entity.getLinkByRel('self');
+	return  entity.href || (self && self.href);
+}
--- a/state/observable/ObserverMap.js
+++ b/state/observable/ObserverMap.js
@@ -1,18 +1,32 @@
+import { deepCopy } from '../../helper/deepCopy.js';
+
 export class ObserverMap {
 	constructor() {
 		this._observers = new Map();
 		this._methods = new WeakMap();
-		this.value = undefined;
 	}
 
 	add(observer, property, method) {
+
+		if (observer === undefined || typeof observer !== 'object') {
+			return;
+		}
+
+		if (property === undefined || typeof property !== 'string') {
+			return;
+		}
+
 		if (this._observers.has(observer)) {
 			return;
 		}
 
-		this.value = value;
 		this._observers.set(observer, property);
-		method && this._methods.set(observer, method);
+
+		if (method) {
+			this._methods.set(observer, method);
+		}
+
+		this._setObserverProperty(observer, property);
 	}
 
 	delete(observer) {
@@ -21,17 +35,21 @@ export class ObserverMap {
 	}
 
 	setProperty(value) {
-		this.value = value;
+		this._value = value;
 
 		this._observers.forEach((property, observer) => {
-			const method = this._methods.has(observer) && this._methods.get(observer);
-			observer[property] = method ? method(this.value) : this.value;
+			this._setObserverProperty(observer, property);
 		});
 	}
 
-	_setObserverProperty(observer) {
+	get value() {
+		return this._value;
+	}
+
+	_setObserverProperty(observer, property) {
 		const method = this._methods.has(observer) && this._methods.get(observer);
-		observer[this._observers.get(observer)] = method ? method(this.value) : this.value;
+		const value = deepCopy(this._value);
+		observer[property] = method ? method(value) : value;
 	}
 
 }
--- a/state/observable/SirenAction.js
+++ b/state/observable/SirenAction.js
@@ -28,7 +28,7 @@ export class SirenAction extends Fetchable(Observable) {
 
 	// TODO: remove in US121366
 	addObserver(observer, property, { method }) {
-		super.addObserver(observer, property, method, this.action);
+		super.addObserver(observer, property, method);
 	}
 
 	get headers() {
--- a/state/observable/SirenAction.js
+++ b/state/observable/SirenAction.js
@@ -11,7 +11,7 @@ export class SirenAction extends Fetchable(Observable) {
 	}
 
 	get action() {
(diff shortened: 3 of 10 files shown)


state/observable/ObserverMap.js
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		this.value = value;
+		this._observers.set(component, property);
+		method && this._methods.set(component, method);
+		this._setObserverProperty(component, value);
+	}
+
+	delete(component) {
+		this._methods.delete(component);
```
zinabat, line 20: What if `method` wasn't passed when the observer was added?

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,43 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = {};
+	}
+
+	add(component, property, method, value) {
```
zinabat, line 8: We should rename all of these `component` parameters to `observer` the same way we did in the `observable` classes

state/observable/SirenAction.js
```diff
@@ -5,27 +5,23 @@ import { refreshToken } from '../token.js';
...
 		super();
-		this._action = { has: false, perform: () => undefined, update: () => undefined };
 		this._name = name;
 		this._token = token;
 		this._state = state;
 	}
 
 	get action() {
-		return this._action;
+		return this._components.value || { has: false, perform: () => undefined, update: () => undefined };
```
zinabat, line 14: We should rename `_components` to `_observers`

test/observerMap.test.js
```diff
@@ -0,0 +1,74 @@
...
+			const value = 'href';
+
+			observerMap.add(component, proprety, method, value);
+
+			expect(observerMap._observers.has(component)).to.be.true;
+			const getProprety = observerMap._observers.get(component);
+			expect(getProprety).to.be.equal(proprety);
+		});
+
+		it('should set value when adding a component', async() => {
```
zinabat, line 28: Keep the language consistent here - use `observer` instead of `component`.

The reason we're making this change is because `observer` isn't necessarily tied to web components.

state/observable/ObserverMap.js
tenbradley: What is the concern?

ciss1995: I believe it just wouldn't do anything? We could also add a check if it exists before doing it

test/observerMap.test.js
ciss1995: Yup good catch

state/observable/SirenSubEntities.js
```diff
@@ -12,18 +12,16 @@ export class SirenSubEntities extends Observable {
...
 		this._token = token;
-		this._entityIds = [];
 	}
 
 	get entityIds() {
-		return this._entityIds;
+		return this._observers.value;
 	}
 
 	set entityIds(entityIds) {
```
tenbradley, line 21: All of the set X functions are nearly identical. Could we alter them to call a function in the superclass that takes in this.X and the value we are setting to reduce redundant code?

state/observable/ObserverMap.js
tenbradley: Calling delete on an element that is not in a Map/WeakMap just returns false. So, we can update it, but this should be sufficient.

state/observable/SirenSubEntities.js
tenbradley: I did this work in https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/17. We don't have to keep it, but this has been addressed.

state/observable/Observable.js
```diff
@@ -1,30 +1,27 @@
...
+import { ObserverMap } from './ObserverMapjs';
+
+export class Observable {
+
+	constructor() {
+		this._observers = new ObserverMap();
+	}
+
+	addObserver(observer, property, { method }, value) {
+		this._observers.add(observer, property, method, value);
```
m6jones, line 10: We shouldn't need `value` parameter anymore, yeah? We may need to double check that is where we are at.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,43 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = {};
+	}
+
+	add(observer, property, method, value) {
```
m6jones, line 8: `value` could be removed from this as well.
We should only be able to set value via SetProperty

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,43 @@
...
+		if (this._observers.has(observer)) {
+			return;
+		}
+
+		this.value = value;
+		this._observers.set(observer, property);
+		method && this._methods.set(observer, method);
+		this._setObserverProperty(observer, value);
```
m6jones, line 16: Value no longer a parameter for this method.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,43 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = {};
```
m6jones, line 5: Should this be left as `undefined` Empty object could see damage I think.

state/observable/SirenAction.js
```diff
@@ -11,24 +11,24 @@ export class SirenAction extends Fetchable(Observable) {
 	}
 
 	get action() {
-		return this._action;
+		return this._observers.value || { has: false, perform: () => undefined, update: () => undefined };
```
m6jones, line 14: I would prefer to store the value in `this._action` not in `this._observers.value` since part of moving the `value` to the `observers` class is so we can disconnect it from the `observer`. 

Basically, in the future we may want to set other properties to value other then `action` Maybe the `href` in this case.

zinabat: Yeah you have it right: Maya is saying whenever the `ObserverMap.value` is set, it should also set the `SirenAction._action` property. It'll be stored in two places, but in this way the `ObserverMap` has control of when this property is set.

state/observable/Observable.js
tenbradley: This is correct. This pr adds value to be stored in the observerMap so we should no longer have to pass it like this.

test/observerMap.test.js
```diff
@@ -0,0 +1,74 @@
...
+			};
+			const proprety = 'proprety';
+			const method = undefined;
+			const value = 'href';
+
+			observerMap.add(observer, proprety, method, value);
+
+			expect(observerMap._observers.has(observer)).to.be.true;
+			const getProprety = observerMap._observers.get(observer);
+			expect(getProprety).to.be.equal(proprety);
```
m6jones, line 25: Property lets make sure this is no where else.

m6jones: Just in this files.

state/observable/SirenAction.js
m6jones: I would probably leave this as is for now.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
...
+		this._methods = new WeakMap();
+		this.value = undefined;
+	}
+
+	add(observer, property, method) {
+		if (this._observers.has(observer)) {
+			return;
+		}
+
+		this.value = value;
```
m6jones, line 13: This needs to be removed.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
...
+		this._observers.set(observer, property);
+		method && this._methods.set(observer, method);
```
m6jones, line 15: We should probably switch this to an if.
```
if (method) {
     this._methods.set(observer, method);
}
```

Just no need to be fancy.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = undefined;
+	}
+
+	add(observer, property, method) {
```
m6jones, line 8: We need valid property is a thing.
required:
observer: object
property: string

Optional
method: function

method should be defined as `false` or `null` to show that we intend it to be optional.

m6jones: function docs explaining the variables.

Ie: How we use method.

Should method be called `callback` for easier reading?

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
...
+		if (this._observers.has(observer)) {
+			return;
+		}
+
+		this.value = value;
+		this._observers.set(observer, property);
+		method && this._methods.set(observer, method);
+	}
```
m6jones, line 16: `_setObserverProperty(observer)` needs to be called for the new observer for it to get the cached value.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
...
+		this._methods.delete(observer);
+		return this._observers.delete(observer);
+	}
+
+	setProperty(value) {
+		this.value = value;
+
+		this._observers.forEach((property, observer) => {
+			const method = this._methods.has(observer) && this._methods.get(observer);
+			observer[property] = method ? method(this.value) : this.value;
```
m6jones, line 28: Call method: `_setObserverProperty(observer)`

state/observable/SirenAction.js
m6jones, line 14: Note the default object here should be `{has: false, commit: () => undefined}`

m6jones: Maybe make it a constant at the top of the file?

m6jones: The `{has: false, commit: () => undefined}` can be a different story.

state/observable/SirenAction.js
```diff
@@ -11,24 +11,24 @@ export class SirenAction extends Fetchable(Observable) {
...
 	}
 
 	set action({ has, perform, update }) {
 		if (!has || typeof perform !== 'function') {
 			perform = () => undefined;
 			update = () => undefined;
 		}
```
m6jones, line 21: Constant at the top would help

state/observable/SirenAction.js
```diff
@@ -11,24 +11,24 @@ export class SirenAction extends Fetchable(Observable) {
...
-		if (this._action.has !== has || this._action.perform !== perform) {
-			this._components.setProperty({ has, perform, update });
+		if (this.action().has !== has || this.action().perform !== perform) {
```
m6jones, line 22: Action isn't a function. So it would be `this.action.has` and `this.action.perform`

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
+export class ObserverMap {
+	constructor() {
+		this._observers = new Map();
+		this._methods = new WeakMap();
+		this.value = undefined;
```
m6jones, line 5: Lets delete this.

Make `value` a getter for `this._value` and modify `this._value` in the rest of the code. 

We only want people to edit `value` via the `setProperty` method.

state/observable/SirenClasses.js
```diff
@@ -2,19 +2,18 @@ import { Observable } from './Observable.js';
 
 export class SirenClasses extends Observable {
 	get value() {
-		return this._value;
+		return this._observers.value;
 	}
 
 	set value(value) {
-		if (!this._value !== value) {
+		if (!this.value() !== value) {
```
m6jones, line 9: `this.value` not `this.value()`

state/observable/SirenClasses.js
```diff
@@ -2,19 +2,18 @@ import { Observable } from './Observable.js';
 
 export class SirenClasses extends Observable {
 	get value() {
```
m6jones, line 4: Rename to classes :D

state/observable/SirenClasses.js
```diff
@@ -2,19 +2,18 @@ import { Observable } from './Observable.js';
...
 			this._observers.setProperty(value);
 		}
-		this._value = value;
 	}
 
 	// TODO: remove in future
 	addObserver(observer, property, { method }) {
-		super.addObserver(observer, property, method, this._value);
+		super.addObserver(observer, property, method);
 	}
```
m6jones, line 17: Removed.

Note this wouldn't work because you have to `super.addObserver(observer, property, { method });`

state/observable/SirenEntity.js
```diff
@@ -3,19 +3,18 @@ import { Observable } from './Observable.js';
...
 			this._observers.setProperty(sirenEntity);
 		}
-		this._sirenEntity = sirenEntity;
 	}
 
 	// TODO: remove in US121366
 	addObserver(observer, property, { method }) {
-		super.addObserver(observer, property, method, this.sirenEntity);
+		super.addObserver(observer, property, method);
 	}
```
m6jones, line 19: This needs to be removed.

state/observable/ObserverMap.js
```diff
@@ -0,0 +1,42 @@
...
+
+		this._observers.forEach((property, observer) => {
+			const method = this._methods.has(observer) && this._methods.get(observer);
+			observer[property] = method ? method(this.value) : this.value;
+		});
+	}
+
+	_setObserverProperty(observer) {
+		const method = this._methods.has(observer) && this._methods.get(observer);
+		observer[this._observers.get(observer)] = method ? method(this.value) : this.value;
```
m6jones, line 34: method could mutate `this.value` if is an object. So we could deepCopy this if we need too.

state/observable/SirenLink.js
```diff
@@ -11,23 +11,21 @@ export class SirenLink extends Observable {
 	}
 
 	get link() {
```
m6jones, line 13: link should be renamed to href.

state/observable/SirenLink.js
```diff
@@ -11,23 +11,21 @@ export class SirenLink extends Observable {
...
 		}
-		this._link = link;
-
 	}
 
 	// TODO: remove in US121366
 	addObserver(observer, property, { route, method }) {
 		if (route) {
 			this._addRoute(observer, route);
 		} else {
```
m6jones, line 27: Does the else here make sense?

state/observable/SirenProperty.js
```diff
@@ -12,20 +12,18 @@ export class SirenProperty extends Observable {
...
 		}
-		this._value = value;
-
 	}
 
 	// TODO: remove in US121366
 	addObserver(observer, property, { method }) {
-		super.addObserver(observer, property, method, this.value);
+		super.addObserver(observer, property, method);
 	}
```
m6jones, line 27: This can be removed.

state/observable/SirenSubEntities.js
```diff
@@ -12,23 +12,21 @@ export class SirenSubEntities extends Observable {
...
-		if (!this._entityId !== entityIds) {
+		if (!this.entityIds() !== entityIds) {
```
m6jones, line 22: to `this.entityIds` not a method

state/observable/SirenSubEntities.js
```diff
@@ -12,23 +12,21 @@ export class SirenSubEntities extends Observable {
...
 			this._observers.setProperty(entityIds || []);
 		}
-		this._entityIds = entityIds || [];
 	}
 
 	// TODO: remove in US121366
 	addObserver(observer, property, { method }) {
-		super.addObserver(observer, property, { method }, this.entityIds);
+		super.addObserver(observer, property, { method });
 	}
```
m6jones, line 30: remove that.

state/observable/SirenProperty.js
```diff
@@ -12,20 +12,18 @@ export class SirenProperty extends Observable {
...
 
 	get value() {
-		return this._value;
+		return this._observers.value;
 	}
 
 	set value(value) {
-		if (!this._value !== value) {
+		if (!this.value() !== value) {
```
m6jones, line 19: This should `this.value`

state/observable/SirenLink.js
m6jones: Yeah this does make sense.

test/observerMap.test.js
```diff
@@ -0,0 +1,76 @@
+import { expect }  from '@open-wc/testing';
```
tenbradley, line 1: If we're being consistent in the repo, I'd use assert and its various modifiers.

state/observable/SirenAction.js
tenbradley: What do you mean?

state/observable/ObserverMap.js
tenbradley: I agree that function docs would be good. But I think that they are needed across this repo. Should we make a separate story to go through and add those docs?

ciss1995:
@tenbradley I agree, we definitely documentations. We can create a story for it.

state/observable/SirenAction.js
ciss1995: Yeah, I didn't get that too

m6jones (approved):

PR #17, opened 2020-10-29
Added generic setValue method to Observable

tenbradley:
US121364/US121366

After reviewing pull requests and writing tests, I observed that this set method would probably be useful in the Observable class now that value is stored in ObserberMaps.

Changes
- added setValue to Observable
- changed most Observable subclasses to call setValue with appropriate params

### DIFF
--- a/state/observable/SirenClasses.js
+++ b/state/observable/SirenClasses.js
@@ -6,9 +6,7 @@ export class SirenClasses extends Observable {
 	}
 
 	set value(value) {
-		if (!this.value() !== value) {
-			this._observers.setProperty(value);
-		}
+		this.setValue(!this.value(), value);
 	}
 
 	// TODO: remove in future
--- a/state/observable/SirenClasses.js
+++ b/state/observable/SirenClasses.js
@@ -2,14 +2,13 @@ import { Observable } from './Observable.js';
 
 export class SirenClasses extends Observable {
 	get value() {
-		return this._value;
+		return this._observers.value;
 	}
 
 	set value(value) {
-		if (!this._value !== value) {
+		if (!this.value() !== value) {
 			this._observers.setProperty(value);
 		}
-		this._value = value;
 	}
 
 	// TODO: remove in future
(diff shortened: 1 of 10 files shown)


ciss1995 (approved):

state/observable/SirenClasses.js
```diff
@@ -6,9 +6,7 @@ export class SirenClasses extends Observable {
 	}
 
 	set value(value) {
-		if (!this.value() !== value) {
-			this._observers.setProperty(value);
-		}
+		this.setValue(!this.value(), value);
```
ciss1995, line 9: Why !?

tenbradley: I am not sure what you are asking about.

AKobets: 2 questions:
  should it be `this.value` instead of `this.value()` because it has a getter.
  Observable has `setValue(oldValue, newValue)` so it seems that the `set value` should call `setValue(this.value, value)`

state/observable/SirenLink.js
```diff
@@ -15,9 +15,7 @@ export class SirenLink extends Observable {
 	}
 
 	set link(link) {
-		if (!this.link() !== link) {
-			this._observers.setProperty(link && link.href);
-		}
+		this.setValue(!this.link(), link && link.href);
```
AKobets, line 18: `link && link.href` probably returns `href`, so it will pass `href` and not the link into `this.setValue`

state/observable/SirenClasses.js
m6jones: Why are you turning `this.value()` into a boolean?

state/observable/SirenLink.js
m6jones: `!this.link()` would return `false` or `true` which should never equal `link && link.href`

m6jones: We want the `href`. Which actually means, we should rename `this.link`

tenbradley: I wrote the code based on what was present in the original code, but may have edited something incorrectly.

state/observable/SirenClasses.js
tenbradley: Because that's what the if previously did?

PR #18, opened 2020-11-17
Added StateFactory

tenbradley:


state/stateFactory.js
```diff
@@ -0,0 +1,17 @@
+import 'd2l-fetch/d2l-fetch.js';
+import { StateStore } from './stateStore.js';
+
+export class StateFactory {
```
m6jones, line 4: Instead of exporting this class could we export 2 methods that use this in the background?

Does this need to be a class? Or should we just have the get method?

The states will be stored globally in `window.hmcFoundation.state` and the exported methods can access that?

tenbradley: If the StateStore is being stored globally, then making it function/method makes sense and the class can be gotten rid of.

state/stateFactory.js
```diff
@@ -0,0 +1,17 @@
...
+	constructor(fetch) {
+		this._d2lfetch = fetch;
```
m6jones, line 6: We don't need fetch in here :D

state/stateFactory.js
tenbradley: To confirm, we will make stateFactory a function in the HypermediaState file. It'll take in a HypermediaState and will run the code that is currently the get method?

state/HypermediaState.js
```diff
@@ -3,6 +3,16 @@ import { Fetchable } from './Fetchable.js';
...
 
+const store = window.D2L.SirenSdk.StateStore;
+
+export function stateFactory(state) {
+	if (store.has(state.entityID, state.token)) {
+		return store.get(state.entityID, state.token);
+	}
+	const result = fetch(state);
+	store.add(result);
+}
```
m6jones, line 14: Store holds the state not the result.

No need to fetch here.

state/HypermediaState.js
```diff
@@ -3,6 +3,16 @@ import { Fetchable } from './Fetchable.js';
 
 export const observableTypes = ot;
 
+const store = window.D2L.SirenSdk.StateStore;
+
+export function stateFactory(state) {
```
m6jones, line 8: Params should be entityId and Token.

state/HypermediaState.js
m6jones: Ohh, also, we need to return the state we just made!

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
+class StateStore {
+	constructor() {
+		this._states = new Map();
+	}
+
+	add(state) {
+		if (!state || !state.entityId || !state.token.toString) {
```
m6jones, line 7: toString is probably a method.

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
+class StateStore {
+	constructor() {
+		this._states = new Map();
+	}
+
+	add(state) {
```
m6jones, line 6: You will need to call: state.refreshToken();

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!state || !state.entityId || !state.token.toString) {
+			return;
+		}
+
+		this._initContainer(state.entityId, state.token);
+	}
+
+	get(entityID, token) {
```
m6jones, line 14: Refresh token should be called here too :O

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
...
+		const lowerCaseEntityId = entityID.toLowerCase();
+		return this.has(entityID, token) && this._states.get(token.toString()).get(lowerCaseEntityId).keys().next().value;
+	}
+
+	has(entityID, token) {
```
m6jones, line 19: Refresh token again

state/HypermediaState.js
tenbradley: I did realize that. :P

state/HypermediaState.js
```diff
@@ -3,6 +3,8 @@ import { Fetchable } from './Fetchable.js';
 
 export const observableTypes = ot;
 
+const store = window.D2L.SirenSdk.StateStore;
```
m6jones, line 6: You will have to make sure this is defined:
```
window.D2L= window.D2L || {};
window.D2L.SirenSdk = window.D2L.SirenSdk || {};
window.D2L.SirenSdk.StateStore = window.D2L.SirenSdk.StateStore || new StateStore()
```

Also, want to change the name away from `SirenSdk` maybe `foundation`

state/HypermediaState.js
```diff
@@ -122,3 +124,12 @@ export class HypermediaState extends Fetchable(Object) {
 		return sirenComponent;
 	}
 }
+
+export function stateFactory(entityID, token) {
+	if (store.has(entityID, token)) {
+		return store.get(entityID, token);
+	}
+	const result = new HypermediaState(entityID, token);
+	store.add(result);
```
m6jones, line 133: You will have to `await` this add due to the token refresh.

state/HypermediaState.js
```diff
@@ -122,3 +129,12 @@ export class HypermediaState extends Fetchable(Object) {
 		return sirenComponent;
 	}
 }
+
+export async function stateFactory(entityID, token) {
+	if (await store.has(entityID, token)) {
+		return await store.get(entityID, token);
+	}
+	const result = new HypermediaState(entityID, token);
```
m6jones, line 137: nit: result can be called state.

m6jones (approved):

PR #19, opened 2020-11-17
Added StateStore

tenbradley:


ciss1995:
LGTM, I'll just add tests.

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
+class StateStore {
```
m6jones, line 1: needs to be exported

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
...
+	constructor() {
+		this._states = new Map();
+	}
+
+	add(state) {
```
m6jones, line 6: needs to be async and add at the start `await state.tokenRefresh()`

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!state || !state.entityId || !state.token.toString) {
+			return;
+		}
+
+		this._initContainer(state.entityId, state.token);
+	}
+
+	get(entityID, token) {
```
m6jones, line 14: needs to be async and add at the start `await state.tokenRefresh()`

state/stateStore.js
```diff
@@ -0,0 +1,36 @@
...
+		const lowerCaseEntityId = entityID.toLowerCase();
+		return this.has(entityID, token) && this._states.get(token.toString()).get(lowerCaseEntityId).keys().next().value;
+	}
+
+	has(entityID, token) {
```
m6jones, line 19: needs to be async and add at the start `await state.tokenRefresh()`

state/stateStore.js
tenbradley: How can I await the tokenRefresh at the start when I don't have the state yet?

ciss1995 (approved):

PR #21, opened 2020-11-24
feat: FETCH, Actions and MORE!

m6jones:
Tested with polarisdev sitelord site with these changes https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/41


state/Fetchable.js
```diff
@@ -30,17 +30,20 @@ class FetchStatus {
 	/**
 	 * Resolves Promise to true and sets pending status to false
 	 */
-	done() {
+	done(response, error) {
 		if (!this.pending) throw new FetchError('Cannot call done() on a status that is not pending');
-		this._resolver(true);
+		error ? this._rejecter(new FetchError(error)) : this._resolver(response);
 		this.pending = false;
 	}
```
m6jones, line 37: I updated this so the promise will return the response or the error. So we can give the raw response. Giving extra power if it is required.

state/Fetchable.js
```diff
@@ -30,17 +30,20 @@ class FetchStatus {
...
 
 	/**
 	 * Creates a Promise and sets pending status to true
 	 */
 	start() {
-		this.complete = new Promise(resolve => this._resolver = resolve);
+		this.complete = new Promise((resolve, reject) => {
+			this._resolver = resolve;
+			this._rejecter = reject;
+		});
```
m6jones, line 46: Promise will get rejected if failed to fetch.

state/Fetchable.js
```diff
@@ -101,6 +110,8 @@ export const Fetchable = superclass => class extends superclass {
 		return 'GET';
 	}
 
+	onServerResponse() {}
```
m6jones, line 113: Mocking this out since it is how a response will be handled.

state/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+	if (fetchable.fetchStatus.pending) {
+		if (!bypassCache) {
+			return fetchable.fetchStatus.complete;
+		}
+		fetchable.fetchStatus.cancel();
+	}
+
+	const responsePromise = fetchable.fetchStatus.start();
+
+	fetchFromServer(fetchable, bypassCache);
```
m6jones, line 15: I am not awaiting this on purpose. I want to make sure people get the returned promise so they can have custom code on responses.

state/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+
+	responsePromise
+		.then(json => fetchable.onServerResponse(json))
+		.catch(error => fetchable.onServerResponse(null, error));
+
+	return responsePromise;
+}
+
+async function fetchFromServer(fetchable, bypassCache) {
```
m6jones, line 24: I hate this name. Any ideas?

state/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+		fetchable.fetchStatus.cancel();
+	}
+
+	const responsePromise = fetchable.fetchStatus.start();
+
+	fetchFromServer(fetchable, bypassCache);
+
+	responsePromise
+		.then(json => fetchable.onServerResponse(json))
+		.catch(error => fetchable.onServerResponse(null, error));
```
m6jones, line 19: Doing this allows cancel to work.

state/observable/SirenAction.js
```diff
@@ -1,31 +1,35 @@
+import { fetch } from '../fetch.js';
```
m6jones, line 1: There's a lot more to go into this file. But selfless action story will handle that.

m6jones:
Last commit was changing `entityId` to `entityID` - notice Ten was using entityID so I just switched it all. entityID is more accurate.

state/HypermediaState.js
```diff
@@ -33,20 +33,17 @@ export class HypermediaState extends Fetchable(Object) {
...
-			sirenComponent.addComponent(component, name, { route: basicInfo.route ? { [name]: basicInfo.route } : undefined, method: observables[name].method });
+			sirenComponent.addObserver(component, name, { route: basicInfo.route ? { [name]: basicInfo.route } : undefined, method: observables[name].method });
 		});
 	}
 
-	createChildState(entityID) {
-		return stateFactory(entityID, this.token.rawToken);
-	}
-
-	createChildStateByRawSirenEntity(rawEntity, token) {
```
m6jones, line 48: Removed the need for this!! Basically, now one can `createChildState` then use `state.setSirenEntity` to load up a entity.

state/HypermediaState.js
```diff
@@ -33,20 +33,17 @@ export class HypermediaState extends Fetchable(Object) {
...
-		return stateFactoryByRawSirenEntity(rawEntity, token);
+	createChildState(entityID, token) {
+		token = token === undefined ? this.token.rawToken : token;
```
m6jones, line 45: Well, sometimes we do need to set the token... Found this out the hard way lol.

state/observable/Observable.js
```diff
@@ -2,16 +2,15 @@ import { ObserverMap } from './ObserverMap.js';
...
-		this._state = state;
 	}
 
-	addObserver(observer, property, { method }) {
+	addObserver(observer, property, { method } = {}) {
 		this._observers.add(observer, property, method);
 	}
 
-	async createChildState(entityID, token) {
+	createChildState(entityID, token) {
```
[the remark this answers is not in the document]
m6jones: Will change back if people want since it might be nice to know how to work with this method.

state/observable/ObserverMap.js
```diff
@@ -26,7 +26,9 @@ export class ObserverMap {
 			this._methods.set(observer, method);
 		}
 
-		this._setObserverProperty(observer, property);
+		if (this._value !== undefined) {
+			this._setObserverProperty(observer, property);
+		}
```
m6jones, line 31: If the value is `undefined` there is no reason to set it on add.

state/observable/SirenLink.js
```diff
@@ -14,15 +16,14 @@ export class SirenLink extends Observable {
 	}
 
 	set href(href) {
-		if (!this.href !== href) {
-			this._observers.setProperty(href && href.href);
+		if (this.href !== href) {
+			this._observers.setProperty(href);
 		}
```
m6jones, line 21: this is now the `href` not the `link`

state/observable/SirenLink.js
```diff
@@ -45,11 +46,13 @@ export class SirenLink extends Observable {
...
+		if (!link) return;
+
+		this.href = link.href;
 
 		if (linkCollectionMap && linkCollectionMap instanceof Map) {
-			this.href.rel.forEach(rel => {
+			link.rel.forEach(rel => {
 				if (linkCollectionMap.has(rel) && this !== linkCollectionMap.get(rel)) {
 					this._merge(linkCollectionMap.get(rel));
 				}
```
m6jones, line 58: This method has changed because it assumed `this.href` used to be called `this.link` was indeed a link. Well, it isn't we changed it to `href` so lets update this method to actually work!

state/observable/SirenLink.js
```diff
@@ -58,12 +61,12 @@ export class SirenLink extends Observable {
...
 		if (this._token) {
-			this._childState = await this.createChildState(this.href.href, shouldAttachToken(this._token, this.href));
+			this._childState = await this.createChildState(link.href, shouldAttachToken(this._token.rawToken, link));
 			this._routes.forEach((route, observer) => {
 				this._childState.addObservables(observer, route);
 			});
-			// TODO: waiting for fetch function
-			//fetch(this._childState);
+
+			fetch(this._childState);
```
m6jones, line 69: The fetch method exists NOW!

state/observable/SirenSubEntities.js
```diff
@@ -1,4 +1,4 @@
-import { getEntityIdFromSirenEntity } from './Common.js';
+import { getEntityIDFromSirenEntity } from './ObserverMap.js';
```
m6jones, line 1: This file changes are all that `EntityId` to `EntityID` oh also removing that Common file...

state/observable/SirenSubEntity.js
```diff
@@ -66,23 +68,24 @@ export class SirenSubEntity extends Observable {
...
 	}
 
 	async _setSubEntity(subEntity) {
-		this.entityId = getEntityIdFromSirenEntity(subEntity);
+		this.entityID = getEntityIDFromSirenEntity(subEntity);
 
 		if (this._token) {
-			this._childState = this._state.createChildStateByyRawSirenEntity(subEntity, this._token);
+			this._childState = await this.createChildState(this.entityID, shouldAttachToken(this._token.rawToken, subEntity));
+			this._childState.setSirenEntity(subEntity);
```
m6jones, line 83: circular dependency and removing of the crazy method!

state/observable/sirenObservableFactory.js
```diff
@@ -26,12 +26,13 @@ const observableClasses = Object.freeze({
...
 
-function definedProperty({ observable: type, prime, rel: id, route, token }) {
+function definedProperty({ observable: type, prime, rel: id, route, token, state }) {
 	return {
 		id,
 		route,
 		token: (prime || route) ? token : undefined,
-		type
+		type,
+		state
```
m6jones, line 35: Most of our observers need the state. So add it to the default roaster of input parameters.

state/observable/sirenObservableFactory.js
```diff
@@ -42,7 +43,7 @@ function handleRouting(observerProperties) {
 	return { ...observerProperties, ...currentProperties, route: observerProperties };
 }
 
-export function sirenDefinedProperty(observerProperties, state) {
+export function sirenObserverDefinedProperty(observerProperties, state) {
```
m6jones, line 46: Happy with either name. Up to debate.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
-import { refreshToken } from './token.js';
-
 export class StateStore {
 	constructor() {
 		this._states = new Map();
 	}
 
-	async add(state) {
-		await state.refreshToken();
```
m6jones, line 9: This `await state.refreshToken`. Was cause a lot of problems!!! State store shouldn't have `async` methods. If it does using it is real bad! Consistency of what is loaded is required.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
...
-		if (!state || !state.entityId || !state.token.toString()) {
+	add(state) {
+		if (!state || !state.entityID || !state.token.toString()) {
 			return;
 		}
 
-		this._initContainer(state.entityId, state.token);
```
m6jones, line 14: Simplified and name change.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
...
-	async get(entityID, token) {
-		await refreshToken(token);
-		const lowerCaseEntityId = entityID.toLowerCase();
+	get(entityID, token) {
+		const lowerCaseEntityID = entityID.toLowerCase();
 		const tokenCache = token.toString();
 
-		const state = this.has(entityID, token) && this._states.get(tokenCache).get(lowerCaseEntityId).keys().next().value;
-		state.refreshToken();
+		const state = this.has(entityID, token) && this._states.get(tokenCache).get(lowerCaseEntityID);
```
m6jones, line 19: This is also part of the simplification

state/token.js
```diff
@@ -66,7 +66,8 @@ let gettingToken = null;
 
 export async function getToken(token) {
 	if (typeof (token) !== 'function') {
-		return new Token(token, token);
+		//todo: this is broken used to be new Token(token, token) which wouldn't work
+		return new Token(token, token || token.rawToken);
```
m6jones, line 70: quick fix so added a todo.

m6jones (commented):
I finally found the review button! Less email spam and a little more organized?

state/Fetchable.js
zinabat: Yeah this is exactly what I had in mind for this class so yay

state/fetch.js
zinabat: If we're not exporting this, can this not just be an anonymous function within the `fetch` definition itself? It looks like we only use it once.

state/observable/Observable.js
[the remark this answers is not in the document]
zinabat: This is what function comments are for.
Just put an `@returns Promise<type the promise resolves to>`: https://jsdoc.app/tags-returns.html

state/observable/SirenAction.js
```diff
@@ -41,13 +45,39 @@ export class SirenAction extends Fetchable(Observable) {
...
+	}
+
 	setBodyFromInput(input) {
 		if (this._rawSirenAction.type.indexOf('json') !== -1) {
 			this._body = JSON.stringify({ ...this._fields, ...input });
 		} else if (this.method !== 'GET' && this.method !== 'HEAD') {
 			const formData = new FormData();
 			const fields = { ...this._fields, ...input };
-			Object.keys(fields).forEach((name) => formData.append(name, fields[name]));
+			Object.keys(fields).forEach((name) => {
```
zinabat, line 78: Stylistically I have no preference on this

state/observable/SirenAction.js
```diff
@@ -73,11 +103,9 @@ export class SirenAction extends Fetchable(Observable) {
 
 		this.action = {
 			has: true,
-			perform: (params) => {
-				this.action.refreshToken();
-				return window.D2L.SirenSdk.StateStore.performAction(this.action, params);
-			},
-			update: (observables) => {
+			commit: (observables) => {
```
zinabat, line 106: <3 so much better

state/observable/SirenAction.js
```diff
@@ -106,4 +134,11 @@ export class SirenAction extends Fetchable(Observable) {
 		}
 		return fields;
 	}
+
+	_prepareAction(observables) {
```
zinabat, line 138: shouldn't we have access to the `observables` from the instance's observable map?

state/observable/SirenLink.js
zinabat: wooooooo!

zinabat (commented):
We should probably bump the version in this PR - can you change the PR title to reflect?

state/observable/SirenAction.js
m6jones: This is a bad name. Will change.

state/fetch.js
m6jones: I am not sure I like this method to be an anonymous function. 

Since the `fetch` method is doing something `async` doesn't support with it's promises and is a little more complicated adding `fetchFromServer` method as an anonymous method inside `fetch` would out shine this small detail making it harder to read and follow. Even worse, someone might see it and switch everything to async for consistency! :O (hopefully tests will catch that though)

This is more personal preference so happy to hear why it should be anonymous.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
...
 export class StateStore {
 	constructor() {
 		this._states = new Map();
 	}
 
-	async add(state) {
-		await state.refreshToken();
-		if (!state || !state.entityId || !state.token.toString()) {
+	add(state) {
+		if (!state || !state.entityID || !state.token.toString()) {
```
lloydwaddell, line 7: Is it possible for token to be undefined? this will throw an error if it is.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
...
 			return;
 		}
 
-		this._initContainer(state.entityId, state.token);
+		const tokenMap = this._initTokenMap(state.token);
+		tokenMap.set(state.entityID, state);
```
lloydwaddell, line 12: States with a mixed case entityID will fail not be able to be returned. Get and has both cast entityID to lower case.

state/stateStore.js
```diff
@@ -1,48 +1,40 @@
...
+		const state = this.has(entityID, token) && this._states.get(tokenCache).get(lowerCaseEntityID);
+
 		return state;
 	}
 
-	async has(entityID, token) {
-		await refreshToken(token);
-		const lowerCaseEntityId = entityID.toLowerCase();
+	has(entityID, token) {
+		const lowerCaseEntityID = entityID.toLowerCase();
```
lloydwaddell, line 25: can entity or token ever be null or undefined? get and has methods are just assuming good data atm

state/stateStore.js
m6jones: This might be possible. I'll check and fix it in the constructor.

state/stateStore.js
m6jones: Yeah, if we this gets bad data then it will send an error to the browser saying `entityID` is null and `toLowerCase` can't be called.

If I caught this I would throw the same error. So assuming good data here is (in my opinion) what we should do. Adding checks makes the code more messy and there is no security risk with bad data here.

However, this is strictly in my opinion and convention would say I am wrong. So it probably pretty safe to assume I am wrong. Can someone explain the value it would add by checking?

state/fetch.js
zinabat: My thinking is that keeping them separate doesn't improve readability, since I expect `fetch` to fetch from the server. I don't find this syntax complicated:
```js
function fetch() {
   // stuff
   (async(fetchable, bypassCache) => {
        // fetch from server code
   })();
}
```
If anything, I find having it defined this way more obvious that we're performing an asynchronous action because the function definition as the `async` keyword defined right in it.

Alternatively we could just name it `performServerFetch` so that the names are at least different enough that people don't get confused - I think either option is fine.

state/stateStore.js
zinabat: Yeah, in both those cases you would get something like: `cannot get property "toLowerCase" of undefined/null`, which I think is descriptive enough of an error for devs. Trying to get something out of a `Map` with a null value is not something I would expect, so that default uncaught error is exactly what it should throw IMO

lloydwaddell: It may not be necessary then, I am looking at the file as an individual class. If any potential errors are being handled elsewhere and this is only used internally, then we understand the implications. Then just need to make sure we have tests only those other systems to make sure it doesn't break.

m6jones: So, we did send an empty `entityID` when I first ran this code from a bug. It was good to check.

state/stateStore.js
```diff
@@ -1,48 +1,43 @@
...
 		const tokenCache = token.toString();
 
-		return this._states.has(tokenCache) && this._states.get(tokenCache).has(lowerCaseEntityId);
+		return this._states.has(tokenCache) && this._states.get(tokenCache).has(lowerCaseEntityID);
 	}
 
-	_initContainer(entityId, token) {
-		const lowerCaseEntityId = entityId.toLowerCase();
+	_initTokenMap(token) {
```
AKobets, line 34: in `add` methot it already does toke.toString() and calls `_initTokenMap` with string value.
`const tokenCache = token.toString();` on line 35 sems redundant.

state/stateStore.js
```diff
@@ -1,48 +1,43 @@
...
-	async add(state) {
-		await state.refreshToken();
-		if (!state || !state.entityId || !state.token.toString()) {
+	add(state) {
+		if (!state || !state.entityID || !state.token?.toString()) {
 			return;
 		}
 
-		this._initContainer(state.entityId, state.token);
+		const lowerCaseEntityID = state.entityID.toLowerCase();
```
AKobets, line 11: why we need to support case insensitive entityID ?

state/stateStore.js
m6jones: Oh, that is why the add method didn't do that.

I agree doing it in the add method is better and have the private methods use the toString methods.

state/stateStore.js
m6jones: They are urls and can be inconsistence... Which they shouldn't be but it happens.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@brightspace-hmc/foundation-engine",
-  "version": "0.0.1",
+  "version": "0.0.1s",
```
zinabat, line 3: ?

state/Fetchable.js
```diff
@@ -79,11 +84,17 @@ export const Fetchable = superclass => class extends superclass {
...
+	 * @param links the links the prime in the cache.
+	 */
+	handleCachePriming() {}
+
 	/**
 	 * @returns {Headers} Headers for the request
 	 */
 	get headers() {
-		if (!this._headers) this._initHeaders();
+		this._initHeaders();
```
zinabat, line 97: The extra check prevents us from setting the headers every time we access them. Why was it removed?

zinabat (commented):
looks good overall

package.json
m6jones: A typo it is! Nice.

state/Fetchable.js
m6jones: Right! We actually need to access them each time. Since the token can change and does change. So if we don't update it each time we are sending around old tokens.

zinabat (approved):

PR #23, opened 2020-11-26
StateStore unit tests 

lloydwaddell:
'''Context'''
Adding unit tests for the stateStore
Updated unit tests for observerMap
Fixed a typo in the imports for siren-observables test

https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F441182461856
https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F457961349652

'''Quality'''
I tested my tests


test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
...
+		});
+
+		it('should return true when there is a match', () => {
+			stateStore.add(defaultState);
+			const foundState = stateStore.has(defaultState.entityID, defaultState.token);
+			assert.isTrue(foundState, 'Default state should have matched');
+		});
+
+		it('should ignore case on entityID', () => {
+			stateStore.add(defaultState);
```
AKobets, line 259: should it explicitely upper the case here before adding to the store ?

test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
...
+			const foundState = stateStore.has(defaultState.entityID.toLowerCase(), defaultState.token);
+			assert.isTrue(foundState, 'EntityID should not be case sensitive');
+		});
+
+		it('should only match exact tokens', () => {
+			stateStore.add(defaultState);
+			const foundState = stateStore.has(defaultState.entityID, defaultState.token.toLowerCase());
```
AKobets, line 266: why `toLowerCase`?

lloydwaddell: Tokens are being compared explicitly in the store, a lower case token should not match a mixed case token

test/stateStore.test.js
lloydwaddell: the token is saved as upper case in the state, see lines 6-10

test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
...
+			stateStore.add(state);
+			assert.equal(stateStore._states.size, 0, 'State was added to the store');
+		});
+
+		it('should not add a state without token', () => {
+			const state = {
+				entityID: entityID
+			};
+
+			stateStore.add(state);
```
AKobets, line 37: is there any return value in stateStore.add or exception thrown?
If it is silently failing how consumer will know that state was not added ?

lloydwaddell: There is no return on store.add(). That is a good point , do maps generally have a return on add / set methods? not sure what is being done for error handling outside of the store.

test/stateStore.test.js
m6jones: tokens shouldn't need `toLowerCase`?

test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
...
+		it('should not add a state without ID', () => {
+			const state = {
+				token: token
+			};
+
+			stateStore.add(state);
+			assert.equal(stateStore._states.size, 0, 'State was added to the store');
+		});
+
+		it('should not add a state with empty token', () => {
```
AKobets, line 50: `It should not add ...`  tests could be converted to data driven approach:
```
const testInput = [{ state: {
				token: '111'
			}, description:  'missing Id'},
           { state: {'entityID: 'abc'}, description: 'missing token'}
];
testInput.forEach((input) => {
it(`should not add state when ${description}`, () => {
            stateStore.add(input.state);
			assert.equal(stateStore._states.size, 0, 'State was added to the store');
})
});
```
It will reduce duplicate code

lloydwaddell: that is handy

test/stateStore.test.js
lloydwaddell: maybe I am being too picky :) Is it be possible for two tokens to be created with the same value but different casing?
so, token-one should not match Token-One

test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
+// import { processRawJsonSirenEntity, stateFactory } from '../state/HypermediaState';
+import { assert }  from '@open-wc/testing';
+import { StateStore } from '../state/stateStore.js';
+
```
AKobets, line 4: genaral comment: https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/master/state/stateStore.js#L23 does refreshToken.
What it is and is it possible to test ?

lloydwaddell: If I remember correctly refreshToken is being removed in maya's branch. There is a big change coming with that branch, so best to check code base from there.

test/stateStore.test.js
m6jones: I just know the token class I made handles it all. But yeah case should matter for tokens.

test/stateStore.test.js
m6jones: the set method for maps will return the object.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/set

So, silent fail it is?

test/stateStore.test.js
m6jones: Convention time:

Do we prefer what Alex mentioned or do it like this:
```
[{ state: {
				token: '111'
			}, description:  'missing Id'},
           { state: {'entityID: 'abc'}, description: 'missing token'}
].forEach((input) => {
it(`should not add state when ${description}`, () => {
            stateStore.add(input.state);
			assert.equal(stateStore._states.size, 0, 'State was added to the store');
})
});
```
Same functionality and it has been done both ways in the UI. I don't care which way we do it. What do people find easier to read?

test/stateStore.test.js
m6jones: Yeah, I told Ten to refresh the token. But really it was just adding onto what the stateStore does. We no longer need to do this in the store.

test/stateStore.test.js
lloydwaddell: personally, I think I prefer Alex's way. Just for the case where it could be handy to reuse the array, otherwise you could still have it written both ways in the code. But I don't have strong opinion on either.

zinabat: I've used both in the past. I often prefer Alex's way because the name of the array gives a developer reading it a chance to quickly understand that they are looking at test cases instead of an out of context array.

tenbradley: I think I like Alex's suggestion because it means the list could be used for a test later as well.

test/observerMap.test.js
```diff
@@ -62,15 +61,40 @@ describe('ObserverMap', () => {
...
+
+		const property = 'property';
+		const method = (value) => { return `${value}-changed`; };
+		const newValue = 'newHref';
+
+		observerMap.add(observer, property, method);
+		observerMap.add(observer2, property, method);
+		observerMap.setProperty(newValue);
+
+		assert(observer[property] === `${newValue}-changed`);
```
AKobets, line 96: please use `assert.equals()`

test/stateStore.test.js
m6jones: Cool we all like Alex's solution the most so lets do that :)

lloydwaddell: implemented in latest commit

test/observerMap.test.js
lloydwaddell: fix all asserts in this file

test/stateStore.test.js
```diff
@@ -0,0 +1,271 @@
+// import { processRawJsonSirenEntity, stateFactory } from '../state/HypermediaState';
```
AKobets: I would suggest to place this file under `test/state/` folder to keep structure similar to src

m6jones:
I rebased my branch. I had to it was required! Luckily it was a rebase so there no history to why I needed to. But trust me, if I didn't truly bad things would have happened.

test/state/stateStore.test.js
```diff
@@ -0,0 +1,191 @@
...
+			{ state: undefined, description: 'is undefined' },
+			{ state: {}, description: 'is empty' },
+			{ state: { entityID: entityID }, description: 'is missing token' },
+			{ state: { token: token }, description: 'is missing entityID' },
+			{ state: { entityID: entityID, token: '' }, description: 'has an empty token' },
+			{ state: { entityID: '', token: token }, description: 'has an empty ID' }
+		];
+
+		invalidStateTests.forEach((test) => {
+			it(`should add a state when state ${test.description}`, () => {
```
tenbradley, line 31: Nice clean way to test and lots of attempts to break things.

tenbradley (commented):
Overall, nice coverage. Alex is the test master, so I'll let him approve it.

test/state/stateStore.test.js
```diff
@@ -0,0 +1,191 @@
+import { assert }  from '@open-wc/testing';
+import { StateStore } from '../../state/StateStore.js';
+
+let stateStore;
+const entityID = 'HREF';
+const token = 'TOKEN';
+const entityLowerCase = entityID.toLowerCase();
```
AKobets, line 7: name should be `entityIdLowerCase`

test/state/stateStore.test.js
```diff
@@ -0,0 +1,191 @@
...
+				assert.isFalse(foundState, 'No matches should be found');
+			});
+		});
+
+		it('should handle call when nothing has been added', () => {
+			const foundState = stateStore.get(defaultState.entityID, defaultState.token);
+			assert.isFalse(foundState, 'Should not return anything if there are no states');
+		});
+
+		it('should ignore case on entityID', () => {
```
AKobets, line 107: just to separate test case from `defaultState.entityID` actual case and cover both lower&upper cases:
```
[defaultState.entityID.toLowerCase(), defaultState.entityID.toUpperCase()].forEach((id) => {
           const foundState = stateStore.get(defaultState.entityID.toLowerCase(), defaultState.token);
			assert.equal(foundState, defaultState, 'Matched the wrong state');
}
```

lloydwaddell: added into the latest commit

AKobets (approved):

PR #25, opened 2020-11-30
Tests added for SirenSubEntities

tenbradley:


state/observable/SirenAction.js
```diff
@@ -25,11 +25,11 @@ export class SirenAction extends Fetchable(Observable) {
 		if (!has || typeof commit !== 'function') {
 			commit = () => undefined;
 		}
-		if (this._action.has !== has || this._action.commit !== commit) {
+		if (this.action.has !== has || this.action.commit !== commit) {
```
tenbradley, line 28: Based on set and get action, it seems we should be calling them and storing in ObserverMap instead of in the Action.

tenbradley: Also, this shouldn't be in this PR, but it snuck in.

state/observable/SirenSubEntities.js
```diff
@@ -7,7 +7,7 @@ export class SirenSubEntities extends Observable {
 		return { token };
 	}
 
-	constructor({ id, token, state }) {
+	constructor({ id, token, state } = {}) {
```
tenbradley, line 10: I had done this in a number of constructor, it can be removed, but I was wondering why it was changed.

state/fetch.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// parse a Link header
+//
+// Link:<https://example.org/.meta>; rel=meta
+//
+// var r = parseLinkHeader(xhr.getResponseHeader('Link');
+// r['meta'] outputs https://example.org/.meta
+//
+function parseLinkHeader(links) {
```
AKobets, line 70: Parsing links with regexp looks scary unless it is well tested.

test/observable/entities.js
```diff
@@ -0,0 +1,25 @@
+export const testSubEntities = [`{
```
AKobets, line 1: please put this file under test/data/observable/entities.js or test/observable/data/entities.js

state/fetch.js
tenbradley: @m6jones This is part of your fetch pr.

state/observable/SirenAction.js
tenbradley: @m6jones Does this make sense?

m6jones: Yeah it does :) good sneak!

state/observable/SirenSubEntities.js
m6jones: A mistake! thats why. Nice catch

state/stateStore.js
```diff
@@ -0,0 +1,41 @@
+export class StateStore {
```
m6jones, line 1: I renamed this file to `StateStore.js` It will be pain to change the case :(

test/data/observable/entities.js
```diff
@@ -0,0 +1,25 @@
+export const testSubEntities = [`{
```
m6jones, line 1: These look awesome! nice organization.

test/fetch.test.js
```diff
@@ -0,0 +1,33 @@
+import { expect }  from '@open-wc/testing';
```
m6jones, line 1: This file needs to be removed for tests to work. I removed it in my pr

test/observable/sirenSubEntities.test.js
```diff
@@ -1,2 +1,84 @@
...
+import { assert }  from '@open-wc/testing';
+import SirenParse from 'siren-parser';
+import { SirenSubEntities } from '../../state/observable/SirenSubEntities.js';
+import { SirenSubEntity } from '../../state/observable/SirenSubEntity.js';
+import { testSubEntities } from '../data/observable/entities.js';
+
+describe('subEntities basic methods', () => {
+	beforeEach(() => {
+
+	});
```
m6jones, line 10: If this is empty it can be removed :)

test/observable/sirenSubEntities.test.js
```diff
@@ -1,2 +1,84 @@
...
+
+	it('sirenSubEntites constructed from empty object', () => {
```
m6jones, line 12: Typo: `sirenSubEntities`

state/stateStore.js
tenbradley: I know. It had been changed, but the file is being a jerk.

state/observable/SirenAction.js
```diff
@@ -9,7 +9,7 @@ export class SirenAction extends Fetchable(Observable) {
 		return { id, token };
 	}
 
-	constructor({ id: name, token, state }) {
+	constructor({ id: name, token, state } = {}) {
```
tenbradley, line 12: I've been debating over whether this makes sense. I feel in practice, it doesn't really make sense to create a SirenX from an empty object. This gives us the syntax `new SirenAction()` which we likely don't want?

test/observable/sirenSubEntities.test.js
tenbradley: I make that typo a lot....

test/observable/sirenSubEntities.test.js
tenbradley: Good catch, I had that doing something but eventually realized I didn't need it.

test/observable/sirenSubEntities.test.js
```diff
@@ -1,2 +1,84 @@
...
+		const obj = new SirenSubEntities();
```
m6jones, line 13: Ohh, we might not have added the `= {}` from `constructor({ state, rel, token } = { })` (written from memory) because it might not make sense for an observable to be in a state were some of these are undefined. 

For these to be meaningful, it should always have token, state. rel could be undefined when we add some future features. So the complete empty state might be a break everything error.

m6jones (approved):
These tests are awesome!

test/observable/sirenSubEntities.test.js
tenbradley: It does currently work, but I think we don't want it working as I mentioned here: https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/25#discussion_r533627244
So, I will fix this so the constructor doesn't have = {}
I still like having a test from an empty object just as a double check that everything appears to be working but I could be convinced to get rid of it.

state/stateStore.js
m6jones: I had to delete the file commit. Make the file again and commit. Probably an easy way but git doesn't seem to care about case.

test/observable/sirenSubEntities.test.js
m6jones: If you kept this test it would be to make sure it fails to be created.

state/fetch.js
m6jones: I copied and pasted this from store.
https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-store.js#L440

It isn't parsing URLS it is parsing a string with a url in it.

m6jones: Basically, copy and pasting doesn't make it safer but means we can steal the tests for it as well. 
https://github.com/Brightspace/polymer-siren-behaviors/blob/master/test/entity-store.js#L404

Also, it currently works in production! So at least status quo will be kept.

test/observable/sirenSubEntities.test.js
tenbradley: Hmm, true. In the updated code, I have it calling `new SirenSubEntities({})` which might not be a good test, but makes some sense.

PR #26, opened 2020-12-01
test: Added coverage configuration

zinabat:
## Context
[US121576](https://rally1.rallydev.com/#/detail/userstory/447027030140?fdp=true)
Fun fact: `open-wc/testing-karma` comes with `istanbul` coverage built in. You just have to pass a flag.

As expected, our code coverage sucks right now, but now we can measure how badly it sucks 🥇 

Screen Shot 2020-12-01 at 1 14 56 PM
Screen Shot 2020-12-01 at 1 04 22 PM

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,4 @@
 {
-    "extends": "brightspace/lit-config"
-  }
-
-  
+  "extends": "brightspace/lit-config",
+  "ignorePatterns": ["coverage/**/*"]
+}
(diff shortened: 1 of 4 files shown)


m6jones (approved):
This is awesome and super happy you checked and notice that we don't need to add another dependancy! 

Wins all around!

m6jones:
Oh, I also love the console output for the test coverage. Simple and has all the information we need. Not overly in your face.

PR #28, opened 2020-12-02
refactor: Replacing sirenComponent with sirenObservable for consistency

zinabat:
## Context
Somehow this didn't happen. Fixing it.

This is also an example of a small PR for you folks who don't remember what they look like 😂 

## Quality
- tested with unit tests
- tested in live LMS environment with bsi

### DIFF
--- a/state/HypermediaState.js
+++ b/state/HypermediaState.js
@@ -25,7 +25,7 @@ export class HypermediaState extends Fetchable(Object) {
 		this._decodedEntity = new Map();
 	}
 
-	addObservables(component, observables) {
+	addObservables(observer, observables) {
 		Object.keys(observables).forEach((name) => {
 			const propertyInfo = {
 				name,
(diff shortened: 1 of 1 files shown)


state/HypermediaState.js
```diff
@@ -36,8 +36,8 @@ export class HypermediaState extends Fetchable(Object) {
 			const basicInfo = sirenObserverDefinedProperty(propertyInfo, this);
 			if (!basicInfo) return;
 
-			const sirenComponent = this._getSirenComponent(basicInfo);
-			sirenComponent.addObserver(component, name, { route: basicInfo.route ? { [name]: basicInfo.route } : undefined, method: observables[name].method });
+			const sirenObservable = this._getSirenObservable(basicInfo);
+			sirenObservable.addObserver(component, name, { route: basicInfo.route ? { [name]: basicInfo.route } : undefined, method: observables[name].method });
```
tenbradley, line 40: Should the variable that is the first parameter be renamed to observer as well?

state/HypermediaState.js
```diff
@@ -48,8 +48,8 @@ export class HypermediaState extends Fetchable(Object) {
 
 	dispose(component) {
 		this._decodedEntity.forEach(typeMap => {
-			typeMap.forEach(sirenComponent => {
-				sirenComponent.delete(component);
+			typeMap.forEach(sirenObservable => {
+				sirenObservable.delete(component);
```
tenbradley, line 52: Here too.

tenbradley (approved):

state/HypermediaState.js
zinabat: yes, thanks

PR #31, opened 2020-12-02
feat: SubEntity and SubEntities observables reflect the siren parsed entity

zinabat:
## Context
[US122956](https://rally1.rallydev.com/#/detail/userstory/460074259704?fdp=true)
The `observableTypes.subEntity` and`observableTypes.subEntities` now reflect a value that is an object. For now, this is the siren entity object parsed from `siren-parser`. This gives the developer creating a component access to the subEntities' links, properties, etc. In the future, this will be a friendly facade that will be easy to interact with when building foundation components.

### Changes
Before, the value of a subEntity and the subEntities array reflected only the `href`. This was problematic for both selfless entities and was also non-intuitive to work with and read.

Subentities are still indexed by their `self` links, but this will be altered in the future to support selfless entities.

### Example of new usage
```js
class CustomComponent  extends HypermediaStateMixin(LitElement)  {
    static get properties() {
        return {
            items: { type: Array, observable: observableTypes.subEntities, rel: rels.item, route: [{observable: observableTypes.link, rel: rels.collection}] }
		};
    }

    render() {
         return html`
              <d2l-list>${repeat(this.items, item => item.href, item => html`
                    <d2l-activity-list-item href="${item.href}" .token="${this.token}" draggable key="${item.properties.id}"></d2l-activity-list-item>
               `)}
	      </d2l-list>
        `;
    }
}
```

## Quality
- Tested with local BSI on `polarisdev` with this branch: https://github.com/BrightspaceHypermediaComponents/foundation-components/tree/zina/subentity-objects
- Updated unit tests to reflect the change

### DIFF
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -17,16 +17,15 @@ export class SirenSubEntities extends Observable {
 		this._rel = id;
 		this._entityMap = new Map();
 		this._token = token;
-		this.entities = [];
 	}
 
 	get entities() {
 		return this._observers.value;
 	}
 
-	set entities(entities) {
-		if (this.entities !== entities) {
-			this._observers.setProperty(entities || []);
+	set entities(sirenParsedEntities) {
+		if (this.entities !== sirenParsedEntities) {
+			this._observers.setProperty(sirenParsedEntities || []);
 		}
 	}
 
@@ -38,10 +37,10 @@ export class SirenSubEntities extends Observable {
 		return this._rel;
 	}
 
-	setSirenEntity(sirenEntity) {
-		const subEntities = sirenEntity && sirenEntity.getSubEntitiesByRel(this._rel);
+	setSirenEntity(sirenParsedEntity) {
+		const subEntities = sirenParsedEntity && sirenParsedEntity.getSubEntitiesByRel(this._rel);
 		const entityMap = new Map();
-		const entities = [];
+		const sirenParsedEntities = [];
 
 		// This makes the assumption that the order returned by the collection
 		// matches the prev/next order for each item.
@@ -61,13 +60,13 @@ export class SirenSubEntities extends Observable {
 				subEntity.entity = sirenSubEntity;
 			}
 			entityMap.set(entityID, subEntity);
-			entities.push(subEntity.entity);
+			sirenParsedEntities.push(subEntity.entity);
 		});
 
 		// Clear the old entity map and reset it to the new one
 		this.entityMap.clear();
 		this._entityMap = entityMap;
 
-		this.entities = entities;
+		this.entities = sirenParsedEntities;
 	}
 }
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -2,6 +2,10 @@ import { getEntityIDFromSirenEntity } from './ObserverMap.js';
 import { Observable } from './Observable.js';
 import { SirenSubEntity } from './SirenSubEntity.js';
 
+/**
+ * Observable SirenSubEntities
+ * Reflects back an array of parsed siren entities to any observers
+ */
 export class SirenSubEntities extends Observable {
 	static definedProperty({ token }) {
 		return { token };
@@ -11,23 +15,23 @@ export class SirenSubEntities extends Observable {
 		super();
 		this._state = state;
 		this._rel = id;
-		this._childSubEntities = new Map();
+		this._entityMap = new Map();
 		this._token = token;
-		this.entityIDs = [];
+		this.entities = [];
 	}
 
-	get entityIDs() {
+	get entities() {
 		return this._observers.value;
 	}
 
-	set entityIDs(entityIDs) {
-		if (this.entityIDs !== entityIDs) {
-			this._observers.setProperty(entityIDs || []);
+	set entities(entities) {
+		if (this.entities !== entities) {
+			this._observers.setProperty(entities || []);
 		}
 	}
 
-	get childSubEntities() {
-		return this._childSubEntities;
+	get entityMap() {
+		return this._entityMap;
 	}
 
 	get rel() {
@@ -36,7 +40,7 @@ export class SirenSubEntities extends Observable {
 
 	setSirenEntity(sirenEntity) {
 		const subEntities = sirenEntity && sirenEntity.getSubEntitiesByRel(this._rel);
-		const childSubEntities = new Map();
+		const entityMap = new Map();
 
 		// This makes the assumption that the order returned by the collection
 		// matches the prev/next order for each item.
@@ -47,27 +51,29 @@ export class SirenSubEntities extends Observable {
 		subEntities.forEach((sirenSubEntity) => {
 			const entityID = getEntityIDFromSirenEntity(sirenSubEntity);
 			// If we already set it up why do it again?
-			if (this.childSubEntities.has(entityID)) {
-				childSubEntities.set(entityID, this.childSubEntities.get(entityID));
-				this.childSubEntities.delete(entityID);
+			if (this.entityMap.has(entityID)) {
+				entityMap.set(entityID, this.entityMap.get(entityID));
+				this.entityMap.delete(entityID);
 				return;
 			}
 
 			const subEntity = new SirenSubEntity({ id: this.rel, token: this._token });
-			subEntity.entityID = entityID;
-			childSubEntities.set(entityID, subEntity);
+			// todo: create a facade to make the subEntity easier to work with
+			sirenSubEntity.href = entityID;
+			subEntity.entity = sirenSubEntity;
+			entityMap.set(entityID, subEntity);
 		});
 
 		// These ones are no longer required.
-		this.childSubEntities.clear();
+		this.entityMap.clear();
 
-		this._childSubEntities = childSubEntities;
+		this._entityMap = entityMap;
 
-		const entityIDs = [];
-		this.childSubEntities.forEach((_, entityID) => {
-			entityIDs.push(entityID);
+		const entities = [];
+		this.entityMap.forEach((subEntity) => {
+			entities.push(subEntity.entity);
 		});
 
-		this.entityIDs = entityIDs;
+		this.entities = entities;
 	}
 }
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -20,7 +20,7 @@ export class SirenSubEntities extends Observable {
 	}
 
 	get entities() {
-		return this._observers.value;
+		return this._observers.value || [];
 	}
 
 	set entities(sirenParsedEntities) {
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -41,6 +41,7 @@ export class SirenSubEntities extends Observable {
 	setSirenEntity(sirenEntity) {
 		const subEntities = sirenEntity && sirenEntity.getSubEntitiesByRel(this._rel);
 		const entityMap = new Map();
+		const entities = [];
 
 		// This makes the assumption that the order returned by the collection
 		// matches the prev/next order for each item.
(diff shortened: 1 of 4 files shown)


state/observable/SirenSubEntities.js
```diff
@@ -47,27 +51,29 @@ export class SirenSubEntities extends Observable {
...
+				entityMap.set(entityID, this.entityMap.get(entityID));
+				this.entityMap.delete(entityID);
 				return;
 			}
 
 			const subEntity = new SirenSubEntity({ id: this.rel, token: this._token });
-			subEntity.entityID = entityID;
-			childSubEntities.set(entityID, subEntity);
+			// todo: create a facade to make the subEntity easier to work with
+			sirenSubEntity.href = entityID;
```
zinabat, line 62: I'm adding the `href` to the object here just for ease of use - eventually we'll have the facade and won't need this.

state/observable/SirenSubEntities.js
```diff
@@ -36,7 +40,7 @@ export class SirenSubEntities extends Observable {
 
 	setSirenEntity(sirenEntity) {
 		const subEntities = sirenEntity && sirenEntity.getSubEntitiesByRel(this._rel);
```
AKobets, line 42: question about existing code:  what is type of syrenEntity, where is `getSubEntitiesByRel` implemented ?

AKobets: > question about existing code: what is type of syrenEntity, where is `getSubEntitiesByRel` implemented ?

Fond it in siren-parser. Assumption is that the sirenEntity is entity object instantiated by siren-parser

zinabat: Yup that's right!

state/observable/SirenSubEntities.js
AKobets: Why not calling `sirenSubEntity.entityID = entityID` ?

AKobets: > I'm adding the `href` to the object here just for ease of use - eventually we'll have the facade and won't need this.

why to modify SirenSubentity object prototype here? Whould it be better to add methods that allow to set href and siren parsed entity to the class ?

state/observable/SirenSubEntities.js
```diff
@@ -47,27 +51,29 @@ export class SirenSubEntities extends Observable {
...
 			const subEntity = new SirenSubEntity({ id: this.rel, token: this._token });
-			subEntity.entityID = entityID;
-			childSubEntities.set(entityID, subEntity);
+			// todo: create a facade to make the subEntity easier to work with
+			sirenSubEntity.href = entityID;
+			subEntity.entity = sirenSubEntity;
+			entityMap.set(entityID, subEntity);
 		});
 
 		// These ones are no longer required.
```
AKobets, line 67: `// These ones are no longer required.` - is it obsolete comment ?

state/observable/SirenSubEntities.js
```diff
@@ -47,27 +51,29 @@ export class SirenSubEntities extends Observable {
 		subEntities.forEach((sirenSubEntity) => {
 			const entityID = getEntityIDFromSirenEntity(sirenSubEntity);
 			// If we already set it up why do it again?
-			if (this.childSubEntities.has(entityID)) {
-				childSubEntities.set(entityID, this.childSubEntities.get(entityID));
-				this.childSubEntities.delete(entityID);
+			if (this.entityMap.has(entityID)) {
+				entityMap.set(entityID, this.entityMap.get(entityID));
+				this.entityMap.delete(entityID);
```
AKobets, line 56: is `this.entityMap.delete(entityID);` necessary? On line 68  it is calling `this.entityMap.clear();`

state/observable/SirenSubEntities.js
```diff
@@ -47,27 +51,29 @@ export class SirenSubEntities extends Observable {
...
-		this.childSubEntities.clear();
+		this.entityMap.clear();
 
-		this._childSubEntities = childSubEntities;
+		this._entityMap = entityMap;
 
-		const entityIDs = [];
-		this.childSubEntities.forEach((_, entityID) => {
-			entityIDs.push(entityID);
+		const entities = [];
```
AKobets, line 72: could it do `entities.push(subEntity.entity)` inside foreach loop, lets say after line 64 ?

state/observable/SirenSubEntity.js
```diff
@@ -76,10 +80,11 @@ export class SirenSubEntity extends Observable {
 	}
 
 	async _setSubEntity(subEntity) {
-		this.entityID = getEntityIDFromSirenEntity(subEntity);
+		const entityId = getEntityIDFromSirenEntity(subEntity);
+		this.entity = subEntity;
 
 		if (this._token) {
-			this._childState = await this.createChildState(this.entityID, shouldAttachToken(this._token.rawToken, subEntity));
+			this._childState = await this.createChildState(entityId, shouldAttachToken(this._token.rawToken, subEntity));
```
AKobets, line 87: Question about implementation of Observable class. Its `createChildState` method calls `this._state.createChildState(entityID, token);` However I could not find where `this._state` is initialized in Observable. Coould it be a bug ?

test/observable/sirenSubEntities.test.js
```diff
@@ -59,10 +59,10 @@ describe('sirenSubEntities set sirenEntity', () =>  {
...
 		subentites.setSirenEntity(entity);
 
-		assert.equal(subentites.childSubEntities.size, 2, 'SirenSubEntities should have 2 children');
-		assert.isTrue(subentites.childSubEntities.has('www.def.com'), 'SirenSubEntities should store first child');
-		assert.isTrue(subentites.childSubEntities.has('www.xyz.com'), 'SirenSubEntities should store second child');
-		assert.deepEqual(subentites.entityIDs, ['www.def.com', 'www.xyz.com'], 'SirenSubEntities entityIDs should store two hrefs');
+		assert.equal(subentites.entityMap.size, 2, 'SirenSubEntities should have 2 children');
+		assert.isTrue(subentites.entityMap.has('www.def.com'), 'SirenSubEntities should store first child');
+		assert.isTrue(subentites.entityMap.has('www.xyz.com'), 'SirenSubEntities should store second child');
+		assert.deepEqual(subentites.entities, entity.entities, 'SirenSubEntities entities should store two hrefs');
```
AKobets, line 65: `store two hrefs` are we storing hrefs or entities ?

test/observable/sirenSubEntity.test.js
```diff
@@ -30,13 +30,13 @@ describe('sirenSubEntity method tests', () => {
...
-	it('subEntity constructed from appropriate object and entityID set', () => {
+	it('subEntity constructed from appropriate object and entity set', () => {
 		const sub = new SirenSubEntity({ id: 'foo', token: 'bar' });
-		sub.entityID = 'baz';
+		sub.entity = { href: 'baz' };
 
 		assert.equal(sub.rel, 'foo', 'subEntity rel field initialized incorrectly');
 		assert.equal(sub._token, 'bar', 'subEntity token field initialized incorrectly');
-		assert.equal(sub.entityID, 'baz', 'subEntity rel field set incorrectly');
+		assert.deepEqual(sub.entity, { href: 'baz' }, 'subEntity rel field set incorrectly');
```
AKobets, line 39: `subEntity rel field set incorrectly` - is it correct comment ?

test/observable/sirenSubEntity.test.js
```diff
@@ -30,13 +30,13 @@ describe('sirenSubEntity method tests', () => {
 		assert.equal(sub._routes.size, 0, 'subEntity routes should be empty when constructed');
 	});
 
-	it('subEntity constructed from appropriate object and entityID set', () => {
+	it('subEntity constructed from appropriate object and entity set', () => {
 		const sub = new SirenSubEntity({ id: 'foo', token: 'bar' });
-		sub.entityID = 'baz';
+		sub.entity = { href: 'baz' };
```
AKobets, line 35: what will be value of sub.entityID prop? was assertion replaced by mistake ?

state/observable/SirenSubEntity.js
tenbradley: This is definitely part of the code that is messy and gross.
We were going to store state in the Observable, but this doesn't work well with SirenAction which has a Fetchable Mixin. So we're only storing _state in the subclasses that actually use state.
I'm personally not a huge fan of calling fields that are in subclasses in superclass (it makes my C++ brain hurt) but Javascript is cool with it. 😢 
What could be better would be to have `this._state` in the superclass and have subclasses set the state with a settor in the subclass. This does better reflect that observable have a state and makes this less gross.
At the very least, we should probably have the line in observables have an if condition to not break if _state is not set.

state/observable/SirenSubEntities.js
zinabat: I will remove it.

state/observable/SirenSubEntities.js
zinabat: We're not setting the entity id because for parsed siren objects, the `href` is the entityID - it's more consistent with the language.
In this case, the `SirenSubEntity` is not what we're reflecting, it's the entity object that's attached to it (`SirenSubEntity.entity`). But you're probably right that the setter should by on the `SirenSubEntity.entity` and not adding to the prototype here; I'll adjust.

test/observable/sirenSubEntities.test.js
zinabat: Good catch

test/observable/sirenSubEntity.test.js
zinabat: We used to be reflecting `entityID` only, but we no longer are. This change is intentional

state/observable/SirenSubEntities.js
m6jones: The comment was expressing that we need to clear the resources of the map to prevent memory leaks. The only thing left in the map were entities that would no longer be linked.

state/observable/SirenSubEntity.js
m6jones: I'd be happy with the change Ten suggests.

m6jones (approved):

state/observable/SirenSubEntity.js
tenbradley: That being said, I would say it's not really part of this pr.

state/observable/SirenSubEntities.js
zinabat: I cleaned up this comment.

state/observable/SirenSubEntities.js
zinabat: I rewrote this function keeping this in mind.

state/observable/SirenSubEntities.js
zinabat: I moved this so we set the `href` when we set the `SirenSubEntity.entity`

test/observable/sirenSubEntity.test.js
zinabat: fixed

state/observable/SirenSubEntities.js
```diff
@@ -11,23 +15,23 @@ export class SirenSubEntities extends Observable {
 		super();
 		this._state = state;
 		this._rel = id;
-		this._childSubEntities = new Map();
+		this._entityMap = new Map();
 		this._token = token;
-		this.entityIDs = [];
+		this.entities = [];
```
AKobets, line 20: would array `entities` cause reader's confusion or collision with getter/setter `entities` ?

state/observable/SirenSubEntities.js
```diff
@@ -11,23 +15,23 @@ export class SirenSubEntities extends Observable {
...
 
-	get entityIDs() {
+	get entities() {
 		return this._observers.value;
 	}
 
-	set entityIDs(entityIDs) {
-		if (this.entityIDs !== entityIDs) {
-			this._observers.setProperty(entityIDs || []);
+	set entities(entities) {
```
AKobets, line 27: Funny stats: in three lines the `entities` word is used 5 times.

state/observable/SirenSubEntities.js
```diff
@@ -46,28 +51,23 @@ export class SirenSubEntities extends Observable {
...
-		const entityIDs = [];
-		this.childSubEntities.forEach((_, entityID) => {
-			entityIDs.push(entityID);
-		});
+		// Clear the old entity map and reset it to the new one
+		this.entityMap.clear();
+		this._entityMap = entityMap;
 
-		this.entityIDs = entityIDs;
+		this.entities = entities;
```
AKobets, line 71: Will it call `set entities` or assign to entities array defined in constructor ?

zinabat: it will call `set entities` which is what causes reflection back to the web component.

test/observable/sirenSubEntities.test.js
```diff
@@ -22,9 +22,9 @@ describe('subEntities basic methods', () => {
 		assert.equal(obj._state, 'hello', 'subEntities state should be initialized');
 		assert.equal(obj.rel, 'abc', 'subEntities rel should be initialized');
 		assert.equal(obj._token, '1234', 'subEntities token should be initialized');
-		assert.instanceOf(obj.childSubEntities, Map, 'subEntities children should be a map');
-		assert.isArray(obj.entityIDs, 'subEntities entityIDs should be empty list');
-		assert.isEmpty(obj.entityIDs, 'subEntites enitityIDS should start empty');
+		assert.instanceOf(obj.entityMap, Map, 'subEntities children should be a map');
+		assert.isArray(obj.entities, 'subEntities entities should be empty list');
```
AKobets, line 26: message should be `subEntities entities should be Array`

state/observable/SirenSubEntities.js
zinabat: heh, I can adjust the language

state/observable/SirenSubEntities.js
zinabat: I will adjust for readability

state/observable/SirenSubEntity.js
zinabat: #34

state/observable/SirenSubEntities.js
zinabat: I actually realized we don't really need this line at all.

state/observable/SirenSubEntities.js
zinabat: I removed the constructor line because it's unnecessary

PR #35, opened 2020-12-04
tests for HypermediaState

AKobets:
Covered most of the HipermediaState functionality and it is ready for review.
Istambul stats for code coverage: 100% Statements 84/84; 97.78% Branches 44/45; 100% Functions 32/32; 100% Lines 75/75

ToDo: 
   discuss better name for state.createChildState ( not in this PR )
   add test for `prime` parameter when adding observables ( will do at next PR )

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "eslint-plugin-html": "^6.1.0",
     "eslint-plugin-lit": "^1.2.2",
     "eslint-plugin-sort-class-members": "^1.8.0",
+    "fetch-mock": "^9.11.0",
     "karma-sauce-launcher": "^4.1.5",
     "sinon": "^9.2.1"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@brightspace-hmc/foundation-engine",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "engine for foundation hypermedia components",
   "scripts": {
     "lint": "eslint . --ext .js,.html",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@brightspace-hmc/foundation-engine",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "description": "engine for foundation hypermedia components",
   "scripts": {
     "lint": "eslint . --ext .js,.html",
--- a/state/HypermediaState.js
+++ b/state/HypermediaState.js
@@ -46,8 +46,11 @@ export class HypermediaState extends Fetchable(Object) {
 	}
 
 	dispose(observer) {
+		console.log('+++++++++++++++++++++++++++++++++');
+		console.log(JSON.stringify(this._decodedEntity));
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
+				console.log(JSON.stringify(sirenObservable));
 				sirenObservable.delete(observer);
 			});
 		});
@@ -81,6 +84,7 @@ export class HypermediaState extends Fetchable(Object) {
 	}
 
 	push() {
+		// should it be async and await for action.push()? SyrenAction class push method id async.
 		this._childStates().forEach(childState => childState.push());
 		const actions = this._getMap(this._decodedEntity, observableTypes.action);
 		actions.forEach(action => action.push());
@@ -100,6 +104,10 @@ export class HypermediaState extends Fetchable(Object) {
 		this._entity = entity !== null ? entity : this._entity;
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
+				console.log('######################################');
+				console.log(this._entity);
+				console.log(typeMap);
+				console.log('######################################');
 				sirenObservable.setSirenEntity(this._entity, typeMap);
 			});
 		});
@@ -114,11 +122,16 @@ export class HypermediaState extends Fetchable(Object) {
 				...observables[name]
 			};
 
-			const basicInfo = sirenObserverDefinedProperty(propertyInfo);
+			const basicInfo = sirenObserverDefinedProperty(propertyInfo); // should it be like in addObservables:  sirenObserverDefinedProperty(propertyInfo, this);
 			if (!basicInfo) return;
 
 			const sirenObservable = this._getSirenObservable(basicInfo);
-			sirenObservable && (sirenObservable.value = propertyInfo.value);
+			console.log('>>>>>>>>>>>>>>>>>>>>>>>>>>>>');
+			console.log(basicInfo); // Object{id: undefined, route: undefined, token: undefined, type: 3, state: undefined}
+			console.log(sirenObservable); // SirenClasses{_observers: ObserverMap{_observers: Map{}, _methods: WeakMap{}}}
+			console.log(sirenObservable.value); // undefined
+			console.log(propertyInfo.value); // undefined
+			sirenObservable && (sirenObservable.value = propertyInfo.value); // does: x.undefined = y.undefined
 		});
 	}
 
--- a/state/HypermediaState.js
+++ b/state/HypermediaState.js
@@ -33,7 +33,7 @@ export class HypermediaState extends Fetchable(Object) {
 				token: this.token,
 				...observables[name]
 			};
-			console.log(`addObservables: name: ${name}; propertyInfo: ${JSON.stringify(propertyInfo)}`);
+			//console.log(`addObservables: name: ${name}; propertyInfo: ${JSON.stringify(propertyInfo)}`);
 			const basicInfo = sirenObserverDefinedProperty(propertyInfo, this);
 			console.log(basicInfo);
 			if (!basicInfo) return;
@@ -49,11 +49,11 @@ export class HypermediaState extends Fetchable(Object) {
 	}
 
 	dispose(observer) {
-		console.log('+++++++++++++++++++++++++++++++++');
-		console.log(JSON.stringify(this._decodedEntity));
+		//console.log('+++++++++++++++++++++++++++++++++');
+		//console.log(JSON.stringify(this._decodedEntity));
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
-				console.log(JSON.stringify(sirenObservable));
+				//console.log(JSON.stringify(sirenObservable));
 				sirenObservable.delete(observer);
 			});
 		});
@@ -75,7 +75,7 @@ export class HypermediaState extends Fetchable(Object) {
 	}
 
 	async onServerResponse(response, error) {
-		console.log(`onServerResponse: ${JSON.stringify(response)}`);
+		//console.log(`onServerResponse: ${JSON.stringify(response)}`);
 		if (error) throw new FetchError(error);
 
 		const entity = await SirenParse(response);
@@ -89,11 +89,11 @@ export class HypermediaState extends Fetchable(Object) {
 
 	push() {
 		// should it be async and await for action.push()? SyrenAction class push method id async.
-		console.log('---------------------------------------------------');
-		console.log(`push childstates: ${JSON.stringify(this._childStates())}`);
+		////console.log('---------------------------------------------------');
+		//console.log(`push childstates: ${JSON.stringify(this._childStates())}`);
 		this._childStates().forEach(childState => childState.push());
 		const actions = this._getMap(this._decodedEntity, observableTypes.action);
-		console.log(`push actions: ${JSON.stringify(actions)}`);
+		//console.log(`push actions: ${JSON.stringify(actions)}`);
 		actions.forEach(action => action.push());
 	}
 
@@ -105,14 +105,14 @@ export class HypermediaState extends Fetchable(Object) {
 	}
 
 	setSirenEntity(entity = null) {
-		console.log(`setSirenEntity called with ${JSON.stringify(entity)}`);
+		//console.log(`setSirenEntity called with ${JSON.stringify(entity)}`);
 		if (entity && entity.href) {
 			return;
 		}
 		this._entity = entity !== null ? entity : this._entity;
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
-				console.log(`sirenObservable: ${sirenObservable.constructor.name}, typeMap: ${JSON.stringify(typeMap)}, entity: ${JSON.stringify(this._entity)}`);
+				//console.log(`sirenObservable: ${sirenObservable.constructor.name}, typeMap: ${JSON.stringify(typeMap)}, entity: ${JSON.stringify(this._entity)}`);
 				sirenObservable.setSirenEntity(this._entity, typeMap);
 			});
 		});
@@ -131,11 +131,11 @@ export class HypermediaState extends Fetchable(Object) {
 			if (!basicInfo) return;
 
 			const sirenObservable = this._getSirenObservable(basicInfo);
-			console.log('>>>>>>>>>>>>>>>>>>>>>>>>>>>>');
-			console.log(basicInfo); // Object{id: undefined, route: undefined, token: undefined, type: 3, state: undefined}
-			console.log(sirenObservable); // SirenClasses{_observers: ObserverMap{_observers: Map{}, _methods: WeakMap{}}}
-			console.log(sirenObservable.value); // undefined
-			console.log(propertyInfo.value); // undefined
+			//console.log('>>>>>>>>>>>>>>>>>>>>>>>>>>>>');
+			//console.log(basicInfo); // Object{id: undefined, route: undefined, token: undefined, type: 3, state: undefined}
+			//console.log(sirenObservable); // SirenClasses{_observers: ObserverMap{_observers: Map{}, _methods: WeakMap{}}}
+			//console.log(sirenObservable.value); // undefined
+			//console.log(propertyInfo.value); // undefined
 			sirenObservable && (sirenObservable.value = propertyInfo.value); // does: x.undefined = y.undefined
 		});
 	}
@@ -145,9 +145,9 @@ export class HypermediaState extends Fetchable(Object) {
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
 				childStates = [...childStates, ...(sirenObservable.childStates || [])];
-				console.log(`%%% _childStates: ${childStates.length}`);
-				console.log(`%%% sirenObservable.childState: ${sirenObservable.childState}`);
-				console.log(`%%% sirenObservable class name: ${sirenObservable.constructor.name}`);
+				//console.log(`%%% _childStates: ${childStates.length}`);
+				//console.log(`%%% sirenObservable.childState: ${sirenObservable.childState}`);
+				//console.log(`%%% sirenObservable class name: ${sirenObservable.constructor.name}`);
 				sirenObservable.childState && childStates.push(sirenObservable.childState);
 			});
 		});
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "eslint-plugin-html": "^6.1.0",
     "eslint-plugin-lit": "^1.2.2",
     "eslint-plugin-sort-class-members": "^1.8.0",
+    "fetch-mock": "^9.11.0",
```
m6jones, line 29: I dont see why this is required. Mocking d2l-fetch is all we should do. No need to add dependency.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
+
+import { HypermediaState, stateFactory, dispose, processRawJsonSirenEntity } from '../../state/HypermediaState';
```
m6jones, line 2: We shouldn't have access to HypermediaState class directly. This would clear up some of the broken tests below. Since statefactory insures proper constructed state.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+	it('addObservables - add one', () => {
+		const observable = {
+			classes: { type: Array, observable: observableTypes.classes }
+		};
+		const observer = { classes: [] };
+		const state = new HypermediaState('foo', 'bar');
+		state.addObservables(observer, observable);
+	});
+
+	it('addObservables - add duplicate ???', () => {
```
m6jones, line 30: Adding duplicate will do nothing. Since it will just overwrite with the same property which is expected behavior.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+		const observable = {
+			classes: { type: Array, observable: observableTypes.classes }
+		};
+		const observer = { classes: [] };
+		const state = new HypermediaState('foo', 'bar');
+		state.addObservables(observer, observable);
+		state.addObservables(observer, observable);
+	});
+
+	it('addObservables - no basic info ??', () => {
```
m6jones, line 40: If there is elements in object that don't match our format for observable type then they are ignored. If there are none at all nothing happens. This is expected behavior so it is easy to mold into different frameworks like lit.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+		const observable = {
+			something: { type: Array, different: 'foo' }
+		};
+		const observer = { unexpected: [] };
+		const state = new HypermediaState('foo', 'bar');
+		state.addObservables(observer, observable);
+	});
+
+	it('addObservables - add with route ??', () => {
```
m6jones, line 49: Routing is a neat way to transfers the hypermedia graph and build up child state tree structure within the hypermedia state. 

If a route can't be followed then end user doesn't have knowledge of what the route should hold so nothing should happen.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+		state.addObservables(observer, observable);
+	});
+
+	it('createChildState - how it is a child state when there is no actual object relations?', async() => {
+		const state = new HypermediaState('foo', 'bar');
+		const anotherState = await state.createChildState('anotherFoo', 'anotherBar');
+		assert.equal(anotherState.entityID, 'anotherFoo');
+	});
+
+	it('createChildState - token has rawToken property', async() => {
```
m6jones, line 64: There is a complete graph structure being made that represents hypermedia api. This method however, might not be part of the tree for this state so I agree this method name should be changed. 

This method is a hack because testing frameworks dont like cycles in graphs but our API graph does include cycles.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+	it.skip('can dispose observer', () => {
+		const observable = {
+			classes: { type: Array, observable: observableTypes.classes }
+		};
+		const observer = { classes: [] };
+		const state = new HypermediaState('foo', 'bar');
+		state.addObservables(observer, observable);
+		state.dispose(observer);
+		/* this test is failing.
+		TypeError: sirenObservable.delete is not a function
```
m6jones, line 79: Seems like there a bug here so yay! I might have called delete over dispose. Oops.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+			at Map.forEach (<anonymous>)
+			at state/HypermediaState.js:2647:15
+			at Map.forEach (<anonymous>)
+			at HypermediaState.dispose (state/HypermediaState.js:2644:25)
+			at Context.<anonymous> (test/state/HypermediaState.test.js:68:9)
+			at Test.Runnable.run (node_modules/karma-mocha-snapshot/lib/adapter.js:14:17)
+		*/
+	});
+
+	it.skip('handleCachePriming - why it is part of the class instance methods, it should be a class static method', async() => {
```
m6jones, line 90: We could make this static. But then we would have to pass the instances token in when we use it. This is an override from the abstract fetchable class.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+		const state = new HypermediaState('foo', 'bar');
+		await state.handleCachePriming(['here', 'there']);
+		/*
+		TypeError: Cannot read property 'rawToken' of undefined
+        at getToken (state/token.js:1284:97)
```
m6jones, line 95: This error is coming from a malformed hypermedia state. The caused by use exporting the raw hypermedia state.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,243 @@
...
+
+	it('handleCachePriming - when token has rawTokenProperty', async() => {
+		const mockLink = fetchMock
+			.mock('path:/link1', {})
+			.mock('path:/link2', {});
+		const state = new HypermediaState('foo', { rawToken: 'bar' });
+		await state.handleCachePriming(['/link1', '/link2']);
+		assert.isTrue(mockLink.called('path:/link1'));
+		assert.isTrue(mockLink.called('path:/link2'));
+		/* implementation has promise.All - dangerous, it is an accidental way to overload hypermedia server*/
```
m6jones, line 114: This isn't dangerous here since we use proper connection libraries. 

The engine wraps a the implementation of a library connecting with the server. The code that makes the connection to the server should be written to obey and not overload the server they are written to. 

We use d2l-fetch and that should handle it. It is expected that we tell d2l-fetch to grab something from the server and d2l-fetch will then return a promise saying it will tell us when it has the data. We don't and shouldn't care how it does that as long as it does it. 

On d2l supported browsers there is a limit of 6 concurrent connections. The hypermedia proxy has dos protection so we are covered for our end users not getting errors and our server not failing.

m6jones (commented):
Here some answers to your questions. My weekend has been pretty bad - so my mathy part took over so a little more blunt. 

It looks good and a great start!

test/state/HypermediaState.test.js
AKobets: thanks, will remove the comment

test/state/HypermediaState.test.js
AKobets: > This error is coming from a malformed hypermedia state. The caused by use exporting the raw hypermedia state.

fixed in the test

test/state/HypermediaState.test.js
AKobets: bug is already fixed.

package.json
AKobets: fetch-mock is a nice library that easy to use. And it is a dev-dependency

state/fetch.js
```diff
@@ -2,7 +2,7 @@ import 'd2l-fetch/d2l-fetch.js';
 
 const d2lfetch = window.d2lfetch;
 
-export function fetch(fetchable, bypassCache = false) {
+export async function fetch(fetchable, bypassCache = false) {
```
m6jones, line 5: This shouldn't be async

state/fetch.js
```diff
@@ -12,7 +12,7 @@ export function fetch(fetchable, bypassCache = false) {
 
 	const responsePromise = fetchable.fetchStatus.start();
 
-	performServerFetch(fetchable, bypassCache);
+	await performServerFetch(fetchable, bypassCache); // needed to add await to see the test working properly.
```
m6jones, line 15: We can't await this.

m6jones: You also don't need to await this since we are returning another promise that tracks if the fetch succeeds or not. 
So you can still write `await fetch()` without this being async and awaiting this.

AKobets: awaiting for fetch does not work if we do not `await performServerFetch`
If in browser we don't want `await performServerFetch` then I will reimplement a test to wait for expected state instead of asserting the state right after the call to fetch.

m6jones: So we can await fetch as shown in my PR here: https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/42

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,717 @@
...
+			const observer = {};
+			const state = new HypermediaState('foo', 'bar');
+			state.addObservables(observer, observable);
+			state.setSirenEntity(SirenParse(JSON.stringify(entity)));
+			assert.deepEqual(observer, { classes: entity.class });
+			state.setSirenEntity(SirenParse(JSON.stringify(updatedEntity)));
+			assert.deepEqual(observer, { classes: updatedEntity.class });
+		});
+
+		it('setSirentEntity - why could not continue if entity has href ?', async() => {
```
m6jones, line 237: An entity with an href means it hasn't been fetched yet. So this says dont set an entity that hasn't been fetched.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,717 @@
...
+	});
+
+	describe('push and reset methods', () => {
+		const methods = ['push', 'reset'];
+		let state, putActionSpy, getActionSpy, linkChildStateSpy, subEntityChilStateSpy;
+		before(async() => {
+			const entityHref = `http://entity-${uniqueId()}`;
+			const observable = {
+				actionPut: { observable: observableTypes.action, name: 'do-put' },
+				// why need route to pass token to observable ?. Without token link does not create a child state
```
m6jones, line 362: Links are not always a link to hypermedia state. Example link to course homepage.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,717 @@
...
+			state.addObservables(observer, observables);
+			const observablObj = state._getSirenObservable({ type: observableTypes.property, id: 'description' });
+			assert.isUndefined(observablObj.value);
+			observables.description.value = 'abc';
+			state.updateProperties(observables);
+			assert.equal(observablObj.value, 'abc');
+			assert.deepEqual(observer, { description: 'abc' });
+
+			// this method needs comment in the code about how it is intended to be used.
+			// My observation:
```
m6jones, line 450: This is what it does. It allows the frontend to predict the server response for a better UX in some situations.

m6jones: It probably needs a better name. Like updateObservers?

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,717 @@
...
+			state = new HypermediaState(entityHref, { rawToken: 'token' });
+			const observer = {};
+
+			state.addObservables(observer, observable);
+			state.setSirenEntity(SirenParse(JSON.stringify(entity)));
+			/*
+				state.setSirenEntity does not await for observables.setSirenEntity
+				and the SirenLink.childState is not created yet because it is calling async method createChildState
+				Solution that I don't like: put delay before continuing.
+				Solution that someone may not like: await for observables to finish with setSirenEntity.
```
m6jones, line 400: The 3rd solution that we are implementing fot skeletons will fix it. The solution we are going to use is like a promise queue that I did in the sirenSDK

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,717 @@
...
+		};
+		const observer = { classes: [] };
+		state.addObservables(observer, observable);
+		const spy = sinon.spy(state);
+		dispose(state, observer);
+		assert.isTrue(spy.dispose.called);
+	});
+	it('dispose - question about dispose purpose', () => {
+		// 1. stateFactory is adding state to the StateStore.
+		// When disposing state does it need to remove it from the store ?
```
m6jones, line 634: We only want to clean up the observed objects that are no longer valid. Keeping the state is good to prevent extra calls. Everything is deleted on new page loads

test/state/HypermediaState.test.js
AKobets: > An entity with an href means it hasn't been fetched yet. So this says dont set an entity that hasn't been fetched.

could you please point me to the code that sets href to the entity that hasn't been fetched ?

test/state/HypermediaState.test.js
AKobets: > Links are not always a link to hypermedia state. Example link to course homepage.

do you expect that someone will create observable on non HM pages ?

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,719 @@
...
+			await fetch(state);
+			await (new Promise(resolve => setTimeout(() => resolve(), 200)));
+		} catch (e) {
+			processingError = e;
+		}
+
+		assert.isTrue(mock.called(selfHref));
+		assert.instanceOf(processingError, FetchError, 'should thow FetchError object');
+		assert.isTrue(spy.setSirenEntity.notCalled);
+		assert.isTrue(spy.onServerResponse.called);
```
AKobets, line 715: after reverting fetch change `await performServerFetch(fetchable, bypassCache); ` => `performServerFetch(fetchable, bypassCache);` this assertion starts failing. Does fetch need to call fetchable.call onServerResponse if server responses with error ?

test/state/HypermediaState.test.js
m6jones: It is sent down by the LMS. If you do a 0 depth on the enrollment domain or organization domain you will see these on the subentities.

test/state/HypermediaState.test.js
m6jones: Yeah that will be expected.

Adding `prime` or `route` is only useful on a hypermedia route so that is why the token is added.

(prime is a keyword to start fetching the link as soon as the component loads)

test/state/HypermediaState.test.js
m6jones: We could handle it with another method like `onServerResponseError` or `annouceError` We don't have the designs for how errors will be  prompted to the user so I've been waiting on those before thinking of a good way to handle them. For now using the second parameter in `onServerResponse` was the quick solution to make sure the rest of the code can get the errors to `HypermediaState`

package.json
m6jones: Yeah, I saw Lloyd use it too. Does seem pretty neat.

test/state/HypermediaState.test.js
AKobets: is prime related to handleCachePriming ? I could not find any reference for it in readme
Will need to add tests for prime

test/state/HypermediaState.test.js
AKobets: fixed with latest master

test/state/HypermediaState.test.js
AKobets: > It is sent down by the LMS. If you do a 0 depth on the enrollment domain or organization domain you will see these on the subentities.

created issue for addeng comments to the setSirenEntity https://github.com/BrightspaceHypermediaComponents/foundation-engine/issues/47

AKobets: also added issue to document parameters for observables factory  and provide examples of indented behavior in observbles. https://github.com/BrightspaceHypermediaComponents/foundation-engine/issues/49

state/HypermediaState.js
```diff
@@ -116,17 +115,15 @@ class HypermediaState extends Fetchable(Object) {
...
 			sirenObservable && (sirenObservable.value = propertyInfo.value);
 		});
 	}
 
 	_childStates() {
-		let childStates = [];
+		const childStates = [];
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
-				childStates = [...childStates, ...(sirenObservable.childStates || [])];
```
m6jones, line 129: This line was for subentities. That would have multiple child states.

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,682 @@
...
+			const state = await stateFactory(entityHref, 'token');
+			const mock = fetchMock
+				.mock(entityHref, JSON.stringify(entity))
+				.mock(linkedHref, JSON.stringify(linkedEntity));
+
+			state.addObservables(observer, observable);
+
+			await fetch(state);
+			// need a delay for onServerResponse callback to fetch&process main entity  then fetch&process liked entity
+			await (new Promise(resolve => setTimeout(() => resolve(), 20)));
```
zinabat, line 144: use `aTimeout` https://open-wc.org/docs/testing/helpers/#atimeout

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,682 @@
...
+
+			assert.isTrue(mock.called(entityHref), `should fetch ${entityHref}`);
+			assert.isTrue(mock.called(linkedHref), `should fetch ${linkedHref}`);
+			assert.deepEqual(observer, {
+				mainEntityClass: ['main-entity-class'],
+				linkedEntityProperty: 'linked-entity-name'
+			}, 'class should be observed from main entity and property should be fetched from linked entity');
+		});
+
+		it('observable with method', async() => {
```
zinabat, line 154:
```suggestion
		it('mutates the observed value with method', async() => {
```

test/state/HypermediaState.test.js
```diff
@@ -0,0 +1,682 @@
...
+			const link2 = `http://link2-${uniqueId()}`;
+			const mockLink = fetchMock
+				.mock(link1, 500)
+				.mock(link2, {});
+			const state = await stateFactory(uniqueId(), 'bar');
+			assert.isFalse(window.D2L.Foundation.StateStore.get(link1, 'bar'));
+			assert.isFalse(window.D2L.Foundation.StateStore.get(link2, 'bar'));
+			let processingError;
+			try {
+				await state.handleCachePriming([link1, link2]);
```
tenbradley, line 300: would `assert.throws(await state.handleCachePriming([link1, link2], FetchError);` be equivalent here?
Is there a preference between the two approaches?

state/HypermediaState.js
AKobets: > This line was for subentities. That would have multiple child states.

there is nothing in the repo that  defines childStates  property

test/state/HypermediaState.test.js
AKobets: > would `assert.throws(await state.handleCachePriming([link1, link2], FetchError);` be equivalent here?
> Is there a preference between the two approaches?

Have tried your suggestion. Id will not work for async properly. Need to use https://www.chaijs.com/plugins/chai-as-promised/
Lets do this refactoring in next PR. Created task  [TA149606](https://rally1.rallydev.com/#/357252966636ud/custom/367300408400?detail=%2Ftask%2F463512651344&fdp=true?fdp=true): use chai-as-promised

test/state/HypermediaState.test.js
m6jones: handleCachePrime is the server telling us what to load.
`prime` is the components tell us what to load.

test/state/HypermediaState.test.js
tenbradley: Makes sense. I was mostly just wondering if there was a difference between the two.

zinabat (approved):

m6jones (approved):

PR #38, opened 2020-12-07
fix: HypermediaState delete update to call deleteObserver

tenbradley:
Context:
@lloydwaddell Found this bug while testing foundation-components(?). The HypermediaState delete method was calling delete on an observable, but the observable method is deleteObserver.

Quality:
- @lloydwaddell updated the line in node_modules locally and the error was fixed.
- Reran tests locally to ensure all tests in the engine pass with this update.


state/HypermediaState.js
```diff
@@ -49,7 +49,7 @@ export class HypermediaState extends Fetchable(Object) {
 	dispose(observer) {
 		this._decodedEntity.forEach(typeMap => {
 			typeMap.forEach(sirenObservable => {
-				sirenObservable.delete(observer);
+				sirenObservable.deleteObserver(observer);
```
AKobets, line 52: could you please review if parameter name needs to be changed from observable to observer in `export function dispose(state, observable)` ?

AKobets: I'm writing tests for `HypermediaState` and you don't need to add tests in this PR.

tenbradley: I agree it should be observer. Have updated.

zinabat (approved):

PR #39, opened 2020-12-08
add SirenSummonAction class

tenbradley:
Context:
[US121372](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F441189124868)
To distinguish selfless actions from other actions, the SirenSummonAction class is being introduced.
Added SirenSummonAction to the sirenFactory

Quality:
- wrote basic tests for construction and gettor/settor
- more tests to be added in further prs for this story
- wrote test to ensure factory can create SummonAction


state/observable/Routable.js
```diff
@@ -0,0 +1,37 @@
+
+export const Routable = superclass => class extends superclass {
```
m6jones, line 2: Could we extend observer and rename this routableObserver. 

That would allow addObserver easier to override. :)

tenbradley: Seems legit. Wasn't sure what was the best way for this to occur in js.
I think we would get into weird things there since SirenSummonAction currently inherits from SirenAction which is not Routable.

m6jones: I see, I see why you didn't do that!
Since extending SummonAction(SirenAction) would be really weird!

tenbradley: Yup. This routable thing might not make sense. It just seemed bad to me to have basically the same code in three places. In theory, this would allow SirenLink and SirenSubEntity to be Routable and clean up those classes (and tests) a bit.

m6jones: Maybe, we just inforce the order of extending. Which is a totally bad idea. However, adding a new type or making this routable will happen so infrequent I think it wouldn't be much of a pain point. Add a javadoc that explains it.

state/observable/Routable.js
```diff
@@ -0,0 +1,37 @@
...
+
+	addObserver(observer, property, { route, method } = {}) {
+		if (route) {
+			this._addRoute(observer, route);
+		} else {
+			super.addObserver(observer, property, { method });
+		}
+	}
+
+	// ideally these would be here too
```
m6jones, line 24: I agree these should be here now

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,38 @@
+import { Routable } from 'Routable.js';
+import { SirenAction } from 'SirenAction.js';
+
+export class SirenSummonAction extends Routable(SirenAction) {
+	constructor({ id: name, token, state, prime }) {
+		super({ name, token, state });
+		this._prime = prime;
+	}
+
+	set action({ has, prime, summon }) {
```
m6jones, line 10: prime doesn't need to be here?

state/observable/Routable.js
tenbradley: Ya. The main reason I didn't suggest doing it before was that I figured we wouldn't be adding types often...and then we added a routable thing.

state/observable/SirenSummonAction.js
tenbradley: I don't know. I'm confused by prime so added it in case?

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,38 @@
...
+	push() {}
+
+	_updateAction() {
+		this.action = {
+			prime: this._prime,
+			has: true,
+			summon: (observables) => {
+				this._prepareAction(observables);
+				this._readyToSend = true;
+				return this._state.updateProperties(observables);
```
m6jones, line 34: This should change to fetch and return the fetch promise.
return fetch(this);

state/observable/Routable.js
m6jones: So having the mixin is preferable for maintenance.
Enforcing order of how we add mixins is okay in this case. Just add a javadoc comment that explains it.

state/observable/SirenSummonAction.js
m6jones: I can think of cases where we could use it. But for now we don't need it.

Prime is basically, hey! make sure you call this route right when this is loaded in. We probably don't need this as routing auto primes. If you don't want to prime you will summon.

state/observable/SirenSummonAction.js
tenbradley: Just the return needs to be changed?

state/observable/Routable.js
zinabat: I am pro jsdoc comments.

state/HypermediaState.js
```diff
@@ -157,7 +157,7 @@ class HypermediaState extends Fetchable(Object) {
 export async function processRawJsonSirenEntity(json, rawToken) {
 	const entity = await SirenParse(json);
 	const entityID = entity.hasLinkByRel && entity.hasLinkByRel('self') && entity.getLinkByRel && entity.getLinkByRel('self').href;
-	if (!entityID) return;
+	if (!entityID) return entity;
```
AKobets, line 160: why it returns  is only in case there is no entityID and does not return anything in ther case. What is a use-case ?

state/observable/SirenAction.js
```diff
@@ -51,7 +51,7 @@ export class SirenAction extends Fetchable(Observable) {
 			return;
 		}
 
-		this._state.processRawJsonSirenEntity(json);
+		return this._state.processRawJsonSirenEntity(json);
```
AKobets, line 54: why there is a need for return ?

state/observable/Routable.js
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	set childState(state) {
+		this._childState = state;
+	}
+
+	/**
+	 * @param {Object} observer
+	 * @param {string} property
+	 * @param {} route
```
zinabat, line 18: JsDoc comments are a little weird with destructured parameters. This is the standard:
```suggestion
	 * @param {Object} obj
	 * @param {Object} obj.route - An object containing information on the route to go through
	 * @param {function} obj.method - A function that will mutate the value before setting the property 
```

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,45 @@
...
+		if (!has || typeof summon !== 'function') {
+			summon = () => undefined;
+		}
+		if (this.action.has !== has || this.action.summon !== summon) {
+			this._observers.setProperty({ has, summon });
+		}
+	}
+
+	onServerResponse(json, error) {
+		const entity = super.onServerResponse(json, error);
```
AKobets, line 27: entity might be assigned `undefined` here because of current implementation of underlying calls.
This PR needs more tests, not just constructor tests

state/observable/SirenAction.js
```diff
@@ -51,7 +51,7 @@ export class SirenAction extends Fetchable(Observable) {
 			return;
```
zinabat, line 51: I always find it really weird when functions sometimes return objects and sometimes don't. There are a couple places we're doing it here - how would you feel about return `null` instead of undefined in this case?

state/HypermediaState.js
```diff
@@ -157,7 +157,7 @@ class HypermediaState extends Fetchable(Object) {
...
 	const state = await stateFactory(entityID, rawToken);
 	state.setSirenEntity(entity);
```
zinabat, line 162: Can we just always return the entity? See note below - it's really weird when sometimes we do and sometimes we don't.

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,45 @@
+import { Routable } from './Routable.js';
+import { SirenAction } from './SirenAction.js';
+
+const defaultAction = { has: true, summon: () => undefined };
+
+export class SirenSummonAction extends Routable(SirenAction) {
```
zinabat, line 6: This makes me smile.

state/HypermediaState.js
tenbradley: It's really returning something if the entity is selfless. This was done to allow the calling function to have the parsed json and notify the action's observers. There might be a nicer/cleaner way to do this.

state/observable/Routable.js
tenbradley: This makes sense. Will update.

state/observable/SirenSummonAction.js
tenbradley: I do agree more tests are needed. This story is broken up into three tasks. I was told not to over test it right now and just have some basic tests.

state/HypermediaState.js
tenbradley: Fair enough. This goes with Alex's comment of why are we returning something if it is selfless. Was trying to find a nice way to handle selflessness before introducing  the Summon class. It _should_ be fine to return nothing in the if case.

state/observable/SirenAction.js
zinabat: It returns an entity _sometimes_ which is an addition to this PR, though see my comments that we should ideally be always returning entity.

state/observable/SirenSummonAction.js
zinabat: It might be a good idea to add more just to keep up code coverage - it's usually easier to test classes when we first create them.

state/observable/SirenSummonAction.js
tenbradley: Me too. :)

state/observable/SirenSummonAction.js
tenbradley: This makes some sense. I felt this function could wait to be tested until SirenAction is tested as it is basically just calling those functions.

_updateAction could be tested. I am a bit scared of setting up a test that uses fetch. :S

state/observable/SirenAction.js
tenbradley: I had gone the opposite way and removed the return so it never returns. Either should be fine.

test/observable/routable.test.js
```diff
@@ -0,0 +1,62 @@
+import { assert }  from '@open-wc/testing';
```
tenbradley, line 1: I stole the tests I had written for SirenLink to be tests for the Routable Mixin. Muahaha.

state/observable/Routable.js
```diff
@@ -0,0 +1,54 @@
...
+ * Behaviour interface for objects that are routable
+ * @mixin
+ */
+export const Routable = superclass => class extends superclass {
+	constructor({ id, token, state }) {
+		super({ id, token, state });
+		this._routes = new Map();
+	}
+
+	get childState() {
```
tenbradley, line 12: Based on how childState is used, I assumed it made sense for it to be in Routable. Does it?

state/observable/SirenSummonAction.js
AKobets: After https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/35 we can create integration e2e tests for this action with HypermediaState / observerFactory /Fetch . Ten please add task to your story for automated  integration testing.

test/observable/routable.test.js
```diff
@@ -0,0 +1,62 @@
...
+import { Observable } from '../../state/observable/Observable.js';
+import { Routable } from '../../state/observable/Routable.js';
+
+class Tester extends Routable(Observable) { }
```
AKobets, line 5: please rename class to RoutableObservable

test/observable/routable.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+describe('adding and removing observers', () => {
+	let subentity, obs;
+
+	beforeEach(() => {
+		subentity = new Tester({});
```
AKobets, line 11: please rename subentity to routableObservable

state/observable/Routable.js
```diff
@@ -0,0 +1,54 @@
...
+
+	/**
+	 * @param {Object} observer
+	 * @param {string} property
+	 * @param {Object} obj
+	 * @param {Object} obj.route - An object containing information on the route to go through
+	 * @param {function} obj.method - A function that will mutate the value before setting the property
+	 */
+	addObserver(observer, property, { route, method } = {}) {
+		if (route) {
```
AKobets, line 28: why both route & method case is not supported here ?

state/observable/Routable.js
```diff
@@ -0,0 +1,54 @@
+/**
+ * Routable mixin
+ * Behaviour interface for objects that are routable
+ * @mixin
+ */
+export const Routable = superclass => class extends superclass {
```
AKobets, line 6: Why we need to have this mixin with nice name `Routable` that  basically reimplements `Observable`

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,44 @@
...
+	constructor({ id: name, token, state, prime }) {
+		super({ id: name, token, state });
+		this._prime = prime;
+		this.action = defaultAction;
```
AKobets, line 10: `this.action = defaultAction` is already in the SirenAction super class

test/observable/sirenSummonAction.test.js
```diff
@@ -0,0 +1,21 @@
...
+	it('construction', () => {
+		const summonAction = new SirenSummonAction({ id: 'foo', token: 'abc', prime: true });
+
+		assert.isTrue(summonAction._prime, 'prime set from constructor parameter');
+		assert.equal(summonAction._name, 'foo', 'name set from constructor parameter');
+		assert.equal(summonAction._token, 'abc', 'token set from constructor parameter');
+		assert.instanceOf(summonAction._routes, Map, 'routes initializes to empty map');
+	});
+
+	it('set action', () => {
```
AKobets, line 14: this test should cover all permutations of `has, summon` states before and after setting an action to assure that 
logic in set action is correct and test covers all branches& statements

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,44 @@
+import { Routable } from './Routable.js';
+import { SirenAction } from './SirenAction.js';
+
+const defaultAction = { has: true, summon: () => undefined };
```
AKobets, line 4: should default be ` has: false` ?

test/siren-observables.test.js
```diff
@@ -68,8 +69,16 @@ describe('observerFactory object creation', () => {
 		assert.instanceOf(basicInfo, SirenSubEntity, 'Object created was not a SirenSubEntity');
 	});
 
+	it('ObservableFactory creates SirenSummonAction', () => {
+		const propertyInfo = {
+			type: ot.summonAction
+		};
+		const basicInfo = sirenObservableFactory(propertyInfo);
+		assert.instanceOf(basicInfo, SirenSummonAction, 'Object created was not a SirenSubEntity');
```
AKobets, line 77: typo `SirenSubEntity` => `SirenSummonAction`

test/siren-observables.test.js
```diff
@@ -68,8 +69,16 @@ describe('observerFactory object creation', () => {
 		assert.instanceOf(basicInfo, SirenSubEntity, 'Object created was not a SirenSubEntity');
 	});
 
+	it('ObservableFactory creates SirenSummonAction', () => {
+		const propertyInfo = {
+			type: ot.summonAction
+		};
+		const basicInfo = sirenObservableFactory(propertyInfo);
```
AKobets, line 76: bad name `basicInfo ` - it should be `observable`

state/observable/SirenSummonAction.js
```diff
@@ -0,0 +1,44 @@
...
+
+export class SirenSummonAction extends Routable(SirenAction) {
+	constructor({ id: name, token, state, prime }) {
+		super({ id: name, token, state });
+		this._prime = prime;
```
AKobets, line 9: prime is assigned but never used. Or I am missing something

state/observable/Routable.js
tenbradley: Do you mean a case where both route and method are added? That's a good question @m6jones

state/observable/Routable.js
tenbradley: The Routable stuff in the Observable is going to be removed. Most observables are not routable so we're moving the routable specific stuff to this mixin to reduce redundant code.
Observable, SirenEntity, and SirenLink will all be updated based on this.

Mostly, I decided if I was making another SirenObject routable, I was making a mixin so that it's easier if we add more classes later to also make them routable.

state/observable/SirenSummonAction.js
tenbradley: I've renamed it to defaultSummon. It's gross to rewrite it, but we want the correct attribute, summon, to be used in SummonAction. What we're setting it to is different than in the SirenAction super class.

test/siren-observables.test.js
tenbradley: This is what happens when I steal my old tests and don't look at them closely enough. :/

state/observable/SirenSummonAction.js
tenbradley: I am still confused by prime. I think it will be used when fetching? @zinabat

test/observable/sirenSummonAction.test.js
tenbradley: Will do.

state/HypermediaState.js
tenbradley: I have updated this code to make it always return the entity.

state/observable/Routable.js
tenbradley: I've cleaned up the observables in #50.

state/observable/Routable.js
tenbradley: If you checkout #50, I've cleaned up Observable and Routable (and some subclasses) to do what I described.

state/observable/SirenSummonAction.js
zinabat: I told Ten to just remove it for now, we can add support later

tenbradley: Can confirm and it has been done.

state/observable/Routable.js
m6jones: When adding a route, we won't be adding an observer. So these cases are separate. Just so happens one needs route and one needs method. 

The question that comes to my mind is - is this the proper place to break into the two cases? Should we have done that earlier in the chain and just make these two separate methods?

tenbradley: I feel that suggests that there should be a better name, such as addObject (which is a very gross name) that is more generic since we aren't always adding an observer here.

zinabat (approved):
I think you should merge this and change your code cleanup PRs to target master

PR #42, opened 2020-12-09
test: fetch tests

m6jones:
It grows!
image

Testing the good old fetch method.

Added JavaDocs comments to explain the methods.

### DIFF
--- a/state/fetch.js
+++ b/state/fetch.js
@@ -24,11 +24,15 @@ export function fetch(fetchable, bypassCache = false) {
 
 	performServerFetch(fetchable, bypassCache);
 
-	responsePromise
-		.then(json => fetchable.onServerResponse(json))
-		.catch(error => fetchable.onServerResponse(null, error));
-
-	return responsePromise;
+	return responsePromise
+		.then(json => {
+			fetchable.onServerResponse(json);
+			return json;
+		})
+		.catch(error => {
+			fetchable.onServerResponse(null, error);
+			throw error;
+		});
 }
 
 /**
--- a/state/fetch.js
+++ b/state/fetch.js
@@ -6,7 +6,7 @@ const d2lfetch = window.d2lfetch;
  * This method sends request to the server. It returns the results to fetchable.onServerResponse and the returned promise.
  * This method also handles cachePrimer header from the server with links with the rel, https://api.brightspace.com/rels/cache-primer.
  * A list of the links that need to be primed are sent to fetchable.handleCachePriming method.
- * @param {Fetchable} fetchable Is an object that inherients from the class fetchable. This is the object that will be fetched.
+ * @param {Fetchable} fetchable Is an object that inherits from the class fetchable. This is the object that will be fetched.
  * @param {boolean} bypassCache Default is `false`. If bypassCache is true then it will cancel any pending request
  *                              and send a new request to the server with no-cache headers.
  * @returns {Promise} This promise will resolve to a json object, or null if the request is canceled.
@@ -33,7 +33,7 @@ export function fetch(fetchable, bypassCache = false) {
 
 /**
  * This method is a helper for fetch. It handles the actual sending of the request.
- * @param {Fetchable} fetchable Is an object that inherients from the class fetchable. This is the object that will be fetched.
+ * @param {Fetchable} fetchable Is an object that inherits from the class fetchable. This is the object that will be fetched.
  * @param {boolean} bypassCache Default is `false`. If bypassCache is true then it will cancel any pending request
  *                              and send a new request to the server with no-cache headers.
  * @returns {Promise} This promise will resolve when fetchable is fetched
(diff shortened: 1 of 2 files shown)


test/fetch.test.js
```diff
@@ -0,0 +1,157 @@
...
+		});
+
+		it('can parse no link header', async() => {
+			headerResponse.Link = '';
+			await fetch(fetchable);
+			expect(fetchable.links).deep.to.equal([]);
+		});
+
+		it('can parse a single link header', async() => {
+			headerResponse.Link = '<https://example.org/.meta>; rel="https://api.brightspace.com/rels/cache-primer"; title="previous chapter"';
```
m6jones, line 134: I'm being clever about how this object is passed by reference :O

test/fetch.test.js
```diff
@@ -0,0 +1,160 @@
...
+			expect(error.message).to.equal('400');
+		}
+	});
+
+	it('fetching the same href at the same time will only run one fetch and return the same promise', async() => {
+		//Being super verbose and step wise on purpose!
+		const fetchableOne = new FetchableObject(hrefGoodStatus, token);
+		const promiseOne = fetch(fetchableOne);
+		const promiseTwo = fetch(fetchableOne);
+		expect(promiseOne).to.equal(promiseTwo);
```
m6jones, line 80: Checking if they are the same object.

test/fetch.test.js
```diff
@@ -0,0 +1,160 @@
...
+
+		const responseOne = await promiseOne;
+		const responseTwo = await promiseTwo;
+
+		expect(fetchstub.calledOnce).is.true;
+		expect(responseOne).to.equal(responseTwo);
```
m6jones, line 86: Checking if they are the same object. In the same place in memory.

test/fetch.test.js
```diff
@@ -0,0 +1,160 @@
...
+		expect(response.request.headers.get('Authorization')).to.equal(`Bearer ${token.value}`);
+		expect(response.request.method).to.equal('GET');
+	});
+
+	it('should throw an error on bad request', async() => {
+		const fetchable = new FetchableObject(hrefBadStatus, token);
+		try {
+			await fetch(fetchable);
+		} catch (error) {
+			expect(error.message).to.equal('400');
```
m6jones, line 71: Check type of error to be a `FetchError`

state/fetch.js
```diff
@@ -2,6 +2,16 @@ import 'd2l-fetch/d2l-fetch.js';
 
 const d2lfetch = window.d2lfetch;
 
+/**
+ * This method sends request to the server. It returns the results to fetchable.onServerResponse and the returned promise.
+ * This method also handles cachePrimer header from the server with links with the rel, https://api.brightspace.com/rels/cache-primer.
+ * A list of the links that need to be primed are sent to fetchable.handleCachePriming method.
+ * @param {Fetchable} fetchable Is an object that inherients from the class fetchable. This is the object that will be fetched.
```
AKobets, line 9: typo `inherients`

AKobets:
Missing test for branch 
image

test/fetch.test.js
```diff
@@ -0,0 +1,160 @@
...
+	it('should fetch an fetchable object that exists', async() => {
+		const fetchable = new FetchableObject(hrefGoodStatus, token);
+		const response = await fetch(fetchable);
+		expect(response.request.body).to.be.null;
+		expect(response.request.headers.get('Authorization')).to.equal(`Bearer ${token.value}`);
+		expect(response.request.method).to.equal('GET');
+	});
+
+	it('should throw an error on bad request', async() => {
+		const fetchable = new FetchableObject(hrefBadStatus, token);
```
AKobets, line 67: need assert that fetchable.onServerresponse is called with error

m6jones: Yeah, I saw you test errors and will be going back to that to copy.

m6jones: I also need to test onServerResponse is called with a json object on success too. Missed both cases.

test/fetch.test.js
```diff
@@ -0,0 +1,218 @@
...
+		const responseTwo = await promiseTwo;
+
+		expect(fetchStub.calledTwice).is.true;
+		expect(responseOne).to.be.null;
+		expect(responseTwo.request.body).to.equal('body');
+		expect(responseTwo.request.headers.get('Authorization')).to.equal(`Bearer ${token.value}`);
+		expect(responseTwo.request.method).to.equal('GET');
+	});
+
+	describe('Link Parse Header', () => {
```
zinabat, line 178: it seems like we can convert these tests to data-driven since they're all pretty much the same test with a different case

test/fetch.test.js
```diff
@@ -0,0 +1,218 @@
...
+			json: async() => jsonResponse,
+			headers: {
+				get: (index) => headerResponse[index]
+			}
+		};
+	},
+	[hrefBadStatus]: () => { return { status: 400, ok: false }; }
+};
+
+before(async() => {
```
zinabat, line 48: for me, this way of mocking out `d2lfetch` is easier to read than the mockFetch stuff, but I think that's just personal preference?

test/fetch.test.js
m6jones: Yeah, I think you are right.

AKobets (approved):

test/fetch.test.js
m6jones: I think it is personal preference. I can see mockFetch being more important if we are mocking more complicated situations.

PR #43, opened 2020-12-09
fix: We shouldn't allow HypermediaState to be exported

m6jones:
Title says it all. Fixed the one set of tests that required it.


zinabat (approved):
🎉

PR #44, opened 2020-12-09
Update fetch to be more consistent in calls.

m6jones:
Basically, what we had before if you `await fetch` the call to onServerResponse may run before or may run after.. This will probably cause a lot of headaches so I've changed it so the promise returned isn't the same promise as `fetchable.fetchStatus` however it should resolve as if it was the same one. This time you can be assured that `onServerResponse` has run and was successful.

This also has the added benefits of people being able to catch errors from `onServerResponse` when calling fetch.

I ran this code through my tests in my other branch. Given the change is complex enough. I decided to switch it off to it's own branch. Also, it seems like it could help the other tests in progress so it will hopefully be merged faster than my test one.

### DIFF
--- a/state/fetch.js
+++ b/state/fetch.js
@@ -14,11 +14,15 @@ export function fetch(fetchable, bypassCache = false) {
 
 	performServerFetch(fetchable, bypassCache);
 
-	responsePromise
-		.then(json => fetchable.onServerResponse(json))
-		.catch(error => fetchable.onServerResponse(null, error));
-
-	return responsePromise;
+	return responsePromise
+		.then(json => {
+			fetchable.onServerResponse(json);
+			return json;
+		})
+		.catch(error => {
+			fetchable.onServerResponse(null, error);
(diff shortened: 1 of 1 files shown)


state/fetch.js
```diff
@@ -14,11 +14,15 @@ export function fetch(fetchable, bypassCache = false) {
...
-
-	return responsePromise;
+	return responsePromise
+		.then(json => {
+			fetchable.onServerResponse(json);
+			return json;
+		})
+		.catch(error => {
+			fetchable.onServerResponse(null, error);
+			throw error;
```
m6jones, line 24: Rethrowing will give the desired behaviour of being able to catch errors on fetch. So you can write:
```
try {
     const response = await fetch(fetchable);
} catch (e) {
     console.error('I got the error too!!!');
}
```

This will be important for more fine control for people who have to call fetch on fetchables.

AKobets (approved):

PR #48, opened 2020-12-10
feat: add summon method SirenSummonAction

tenbradley:
Making the SirenSummonAction summonable.

Quality:
Passes integration test.


state/observable/SirenSummonAction.js
```diff
@@ -25,13 +27,24 @@ export class SirenSummonAction extends Routable(SirenAction) {
 	}
 
 	onServerResponse(json, error) {
-		const entity = super.onServerResponse(json, error);
+		super.onServerResponse(json, error);
+		const entity = this._state.processRawJsonSirenEntity(json);
+
 		this.setSirenEntity(entity);
+		return entity;
```
tenbradley, line 34: Return was modified in super class so now we need to process the entity again.

state/observable/SirenSummonAction.js
```diff
@@ -25,13 +27,24 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
 	}
 
 	// overriding superclass push method to do nothing
 	push() {}
 
+	async summon() {
+		// TODO: return SirenFacade when it exists
+		const response = fetch(this);
```
AKobets, line 42: fetch(this) calls onServerResponse, which already does parsing and setting the entity.

state/observable/SirenSummonAction.js
```diff
@@ -25,13 +27,24 @@ export class SirenSummonAction extends Routable(SirenAction) {
 	}
 
 	onServerResponse(json, error) {
-		const entity = super.onServerResponse(json, error);
+		super.onServerResponse(json, error);
+		const entity = this._state.processRawJsonSirenEntity(json);
```
AKobets, line 31: `this._state.processRawJsonSirenEntity(json)` will parse json and will set the stat's entity again which will call setSirenEntity for all observables in the state.
Is it what you would like it to do here ?

state/observable/SirenSummonAction.js
tenbradley: Hmm, the set might not be necessary, but currently the parsed entity is not returned. I tried to return the parsed entity and it seemed that was not what was wanted. So right now, the parsing does have to occur again.

state/observable/SirenSummonAction.js
tenbradley: I believe that is what it is intended to do. If there is an update, we should tell all of our observers about the update, no?

state/observable/SirenSummonAction.js
AKobets: have you tried `await fetch`

state/observable/SirenSummonAction.js
```diff
@@ -26,12 +27,20 @@ export class SirenSummonAction extends Routable(SirenAction) {
 
 	onServerResponse(json, error) {
 		const entity = super.onServerResponse(json, error);
+
 		this.setSirenEntity(entity);
+		return entity;
```
m6jones, line 32: I don't think onServerResponse needs to return anything.

state/observable/SirenSummonAction.js
```diff
@@ -26,12 +27,20 @@ export class SirenSummonAction extends Routable(SirenAction) {
 
 	onServerResponse(json, error) {
 		const entity = super.onServerResponse(json, error);
+
 		this.setSirenEntity(entity);
```
m6jones, line 31: We don't need this as  this._state.processRawJsonSirenEntity(json); does it in super.

tenbradley: I assume you mean setting the siren entity? From what I can tell, it will only set the siren entity if it has a self link, which summon actions shouldn't?

zinabat:
Quick note that this `gets an action routed through a summon action` is failing because of a flaky test

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,44 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+			this.action = { has: false };
+			return;
+		}
+
+		this._rawSirenAction = entity.getActionByName(this._name);
+		this._href = this._rawSirenAction.href;
+		this._fields = this._decodeFields(this._rawSirenAction);
+
+		this._updateAction();
+		if (this._token) {
```
m6jones, line 51: This will always have a token since we need a token for the action.

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,44 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+			this.routedState = await this.createRoutedState(null, shouldAttachToken(this._token.rawToken, entity));
+			this._routes.forEach((route, observer) => {
+
+				this.routedState.addObservables(observer, route);
+			});
```
m6jones, line 56: This part should be in onServerResponse.

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,44 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+
+			await fetch(this);
```
m6jones, line 58: No need to await this fetch.

state/observable/SirenSummonAction.js
m6jones: The thing returned here is what summon will return. So it will be the facade eventually.

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,45 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+	async onServerResponse(json, error) {
+		const entity = super.onServerResponse(json, error);
+		const sirenEntity = await this._state.processRawJsonSirenEntity(json);
+
+		if (this._routes.size > 0) {
+			this.routedState = await this.createRoutedState(null, shouldAttachToken(this._token.rawToken, sirenEntity));
+			this._routes.forEach((route, observer) => {
+				this.routedState.addObservables(observer, route);
+			});
+		}
```
m6jones, line 39: This doesn't need the `if part.

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,45 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+		if (!entity || !entity.hasActionByName(this._name)) {
+			this.action = { has: false };
+			return;
+		}
+
+		this._rawSirenAction = entity.getActionByName(this._name);
+		this._href = this._rawSirenAction.href;
+		this._fields = this._decodeFields(this._rawSirenAction);
+
+		fetch(this);
```
m6jones, line 59: We need to `if` down here.

state/observable/SirenSummonAction.js
```diff
@@ -24,13 +27,45 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
 	}
 
-	onServerResponse(json, error) {
-		return super.onServerResponse(json, error);
+	async onServerResponse(json, error) {
+		const entity = super.onServerResponse(json, error);
+		const sirenEntity = await this._state.processRawJsonSirenEntity(json);
+
+		if (this._routes.size > 0) {
+			this.routedState = await this.createRoutedState(null, shouldAttachToken(this._token.rawToken, sirenEntity));
```
m6jones, line 35: We should always make this `routedState`

m6jones: `null` should be `const entityID = getEntityIDFromSirenEntity(subEntity);`

zinabat (approved):

PR #50, opened 2020-12-10
refactor: changed childState to routedState. SirenLink and SubEntity use mixin Routable 

tenbradley:
Context:
Basically, we made a routable mixin and this is how it has to be used. I have also made the other Siren classes that should use this mixin use it.

This should make testing of some stuff nicer and routable stuff be more consistent.
Technically part of US121372, but more preventing tech debt for later.

Quality:
- SirenLink and SirenSubEntity pass all tests they previously passed.


state/observable/Routable.js
```diff
@@ -9,12 +11,12 @@ export const Routable = superclass => class extends superclass {
 		this._routes = new Map();
 	}
 
-	get childState() {
-		return this._childState;
+	get routedState() {
```
tenbradley, line 14: I could be convinced this is a bad name, but it seems more accurate than childState.

state/observable/Routable.js
```diff
@@ -32,6 +34,10 @@ export const Routable = superclass => class extends superclass {
 		}
 	}
 
+	createRoutedState(entityID, token) {
```
AKobets, line 37: I could not find where `createRoutedState` is used. Did I miss it?

tenbradley: No, I forgot to change a function call which I have now done.
My plan is to use the term routedState instead of childState as I think it better describes what the state is stored in the routable is.

state/observable/SirenSubEntity.js
```diff
@@ -85,15 +63,8 @@ export class SirenSubEntity extends Observable {
...
-			this._childState = await this.createChildState(entityId, shouldAttachToken(this._token.rawToken, subEntity));
-			this._childState.setSirenEntity(subEntity);
-			this._routes.forEach((route, observer) => {
-				this._childState.addObservables(observer, route);
-			});
-
-			fetch(this._childState);
-		}
+		this.updateRoutedState(entityId, subEntity);
+		this.childState.setSirenEntity(subEntity);
```
m6jones, line 67: This call should be in updateRoutedState

state/observable/Routable.js
```diff
@@ -40,14 +46,25 @@ export const Routable = superclass => class extends superclass {
 		}
 	}
 
+	async updateRoutedState(identifier, subEntity) {
+		if (this._token) {
+			this.routedState = await this.createRoutedState(identifier, shouldAttachToken(this._token.rawToken, subEntity));
```
m6jones, line 51: Should be below this line. Hmm, createRoutedState could handle it?

m6jones: this = `routedState.setSirenEntity`
setSirenEntity needs to be called before fetch. Otherwise we get a redundant fetch.

tenbradley: I think this is the same question: do we always want to set the routedState's sirenEntity? If so, will do.

m6jones:
One of the tests are failing because in the `HypermediaState.push` accesses childState which is now renamed.

tenbradley:
> One of the tests are failing because in the `HypermediaState.push` accesses childState which is now renamed.

Zina and I found that and fixed it and the test was still failing.

test/state/HypermediaState.test.js
```diff
@@ -393,18 +393,18 @@ describe('HypermediaState class', () => {
 			await wait(() => subEntityObservable !== undefined);
 			putActionSpy = sinon.spy(putActionObservable);
 			getActionSpy = sinon.spy(getActionObservable);
-			subEntityChilStateSpy = sinon.spy(subEntityObservable.childState);
-			linkPrimeChildStateSpy = sinon.spy(linkPrimeObservable.childState);
+			subEntityRoutedStateSpy = sinon.spy(subEntityObservable.routedState);
+			linkPrimeRoutedStateSpy = sinon.spy(linkPrimeObservable.routedState);
```
tenbradley, line 397: Changing childState to routedState fixed two of the tests that were broken.

state/observable/SirenSubEntity.js
m6jones: Link doesn't need to do it that is why. But a link will have the href and will be skipped anyways. So it is all good to put it here.

state/observable/SirenLink.js
```diff
@@ -1,13 +1,11 @@
-import { fetch } from '../fetch.js';
 import { Observable } from './Observable.js';
-import { shouldAttachToken } from '../token.js';
+import { Routable } from './Routable.js';
 
-export class SirenLink extends Observable {
+export class SirenLink extends Routable(Observable) {
 	constructor({ id, token, state } = {}) {
-		super();
+		super({});
```
m6jones, line 6: Should we be adding  token, id and route here?

tenbradley: To the super call? It shouldn't be needed. Neither Routable nor Observable holds any of that data at the moment.

m6jones: Oh,

state/observable/Routable.js
```diff
@@ -9,12 +11,12 @@ export const Routable = superclass => class extends superclass {
 		this._routes = new Map();
```
m6jones, line 11: This is more for the parameters.

You should give the constructor the parameter, `object = {}` and then do `super(object)` since right now there too much knowledge bleed on one use case for this constructor.

state/observable/Routable.js
```diff
@@ -1,3 +1,5 @@
+import { shouldAttachToken } from '../token.js';
```
m6jones, line 1: Not importing fetch

state/observable/SirenSubEntity.js
tenbradley: I'm probably going to remove the updateRoutedState function because something about it is causing issues, so this is a moot point.

state/observable/Routable.js
tenbradley: Same comment as before, I think I am going to remove this function.

state/observable/SirenLink.js
tenbradley: Oh?

AKobets (approved):

Issue #49, opened 2020-12-10
Document parameters that passed to sirenObservableFactory


AKobets:
We need readme file with documenting parameters usage for properties used to create observables
The readme should have examples and explanation of obserables behaviour depending on existance or special values in those parameters.
Currently it has: `{ observable: type, prime, rel: id, route, token, state }`

https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/master/state/observable/sirenObservableFactory.js#L29

PR #51, opened 2020-12-14
docs: Added readme for observables

zinabat:
## Context
Issue #49 
Adds basic documentation for the different observables. My plan is to add pages for each type with examples, but I wanted to make sure we document `route` and `prime` properly somewhere.

See: https://github.com/BrightspaceHypermediaComponents/foundation-engine/tree/zina/readme/state/observable

I've intentionally not linked this on the main readme because some of the features aren't implemented yet, like subEntity routing. The readme represents what we are building, not necessarily what we have.

### DIFF
--- a/state/observable/README.md
+++ b/state/observable/README.md
@@ -52,14 +52,13 @@ Observables can be passed the `route` parameter in order to step through the Sir
 
 - `Link`: fetches the entity and returns the observable
 - `SubEntity`: fetches the entity if needed and returns the observable
-- `SubEntities`: fetches the entities if needed and returns an array of the observed items
-- `SummonAction`: calls `summon` to get the entity
+- `SummonAction`: calls `summon` to get the entity and returns the observable
 
 Routing can be used to dig as far into the tree as necessary by adding items to the array.
 
 ### Prime
 
-The `prime` parameter will fetch the relevant observable when the component first loads and place it in the `StateStore`. This is useful for linked entities that are needed when the component loads instead of passing the baton to a sub-component, or to fetch more information on a `subEntity`.
+The `prime` parameter will fetch the relevant observable when the component first loads and place it in the `StateStore`. This is useful for related entities that are needed when the component loads instead of waiting to fetch.
 
 - `Link`: fetches the entity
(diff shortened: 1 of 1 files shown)


state/observable/README.md
```diff
@@ -0,0 +1,68 @@
...
+    rel: 'https://api.brightspace.com/rels/richtext-editor-config'
+  }]
+}
+```
+
+Observables can be passed the `route` parameter in order to step through the Siren Hypermedia tree. Adding the `route` parameter to a routable observable will fetch the relevant entity if needed and create an observable on that entity. The component is then added as an observable to that routed entity.
+
+- `Link`: fetches the entity and returns the observable
+- `SubEntity`: fetches the entity if needed and returns the observable
+- `SubEntities`: fetches the entities if needed and returns an array of the observed items
```
tenbradley, line 55: Currently SirenSubEntities isn't routable. Should it be? Having it here tells me it should be routable.

zinabat: I'm iffy on this one, but it seems like it should be? One use case: a component that wants to observe a list of names attached to multiple sub entities.

state/observable/README.md
```diff
@@ -0,0 +1,68 @@
...
+- `SummonAction`: calls `summon` to get the entity
+
+Routing can be used to dig as far into the tree as necessary by adding items to the array.
+
+### Prime
+
+The `prime` parameter will fetch the relevant observable when the component first loads and place it in the `StateStore`. This is useful for linked entities that are needed when the component loads instead of passing the baton to a sub-component, or to fetch more information on a `subEntity`.
```
m6jones, line 62: The main use for prime is that we know we are going to fetch it so why not do it asap. 

Right now it has a side effect of linking the hypermediaState that is defined by the component and the primed child. I hope that I can think of a way to auto-generate these links.

zinabat: I like that idea

state/observable/README.md
zinabat: A note that I have an idea for what routable subEntities should look like, but it's not something we currently need

tenbradley (approved):

PR #52, opened 2020-12-15
Added SirenFacade and unit tests

zinabat:
## Context
[US123074](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F461413871352&fdp=true?fdp=true)

Introduces the `SirenFacade` object, a class designed to be used for reflecting observable Siren entities to components without attaching extraneous information to them. This is the first iteration and has notable items intentionally missing from it.

- Introduces the class, but existing functionality is unchanged. Coming in next PR
- Actions returns just a list of the names of each action. We are still deciding how we want to handle actions attached to subEntities

## Quality
- new unit tests added, which are modelled after styles I liked from both @tenbradley and @AKobets work
  - side note that I think Alex is right and we should start keeping mock data directly in the tests. I find it easier to read.
- coverage for new file is 100%

### DIFF
--- a/state/observable/SirenFacade.js
+++ b/state/observable/SirenFacade.js
@@ -1,0 +1,36 @@
+import { getEntityIDFromSirenEntity } from './ObserverMap.js';
+/**
+ * A class to transform parsed siren entities into easy to use
+ * objects to be reflected and worked with in foundation components.
+ */
+export class SirenFacade {
+	/**
+	 * Creates the facade from a parsed entity
+	 * @param {Entity} sirenParsedEntity - The raw entity as it has been parsed by SirenParser
+	 * @param {Boolean} verbose - Whether to return more or less information
+	 */
+	constructor(sirenParsedEntity, verbose = false) {
+		if (!sirenParsedEntity) return;
+
+		this.href = getEntityIDFromSirenEntity(sirenParsedEntity) || undefined;
+		this.class = sirenParsedEntity.class || [];
+		this.links = sirenParsedEntity.links || [];
+		// todo: We should figure out how we want to do actions - this is just a list of names
+		this.actions = sirenParsedEntity.actions ? sirenParsedEntity.actions.map(x => x.name) : [];
+		// todo: sub entities only go one level currently
+		this.entities = sirenParsedEntity.entities ? sirenParsedEntity.entities.map(x => new SirenFacade(x)) : [];
+		this.properties = sirenParsedEntity.properties || {};
+		if (verbose) {
+			this.rawSirenParsedEntity = sirenParsedEntity;
+		}
+	}
+
+	/**
+	 * Checks if the class name exists on the entity or not
+	 * @param {String} className - Class name to check
+	 * @returns {Boolean}
+	 */
+	hasClass(className) {
+		return !!(this.class.length && this.class.includes(className));
+	}
+}
--- a/state/observable/SirenFacade.js
+++ b/state/observable/SirenFacade.js
@@ -18,7 +18,7 @@ export class SirenFacade {
 		// todo: We should figure out how we want to do actions - this is just a list of names
 		this.actions = sirenParsedEntity.actions ? sirenParsedEntity.actions.map(x => x.name) : [];
 		// todo: sub entities only go one level currently
-		this.entities = sirenParsedEntity.entities ? sirenParsedEntity.entities.map(x => new SirenFacade(x)) : [];
+		this.entities = sirenParsedEntity.entities ? sirenParsedEntity.entities.map(x => new SirenFacade(x, verbose)) : [];
 		this.properties = sirenParsedEntity.properties || {};
 		if (verbose) {
 			this.rawSirenParsedEntity = sirenParsedEntity;
--- a/test/data/observable/entities.js
+++ b/test/data/observable/entities.js
@@ -1,66 +1,66 @@
 export const subEntitiesTests = { fooEntity:`{
-  "entities": [{"rel": ["foo"], "href": "www.abc.com"}],
-  "links": [{
-	"rel": ["self"],
-	"href": "http://example.com"
-  }],
-  "href": "hello"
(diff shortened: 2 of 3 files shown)


test/data/observable/entities.js
```diff
@@ -1,66 +1,66 @@
 export const subEntitiesTests = { fooEntity:`{
```
zinabat, line 1: Editorconfig issue from elsewhere.

state/observable/SirenFacade.js
```diff
@@ -0,0 +1,36 @@
...
+		this.href = getEntityIDFromSirenEntity(sirenParsedEntity) || undefined;
+		this.class = sirenParsedEntity.class || [];
+		this.links = sirenParsedEntity.links || [];
+		// todo: We should figure out how we want to do actions - this is just a list of names
+		this.actions = sirenParsedEntity.actions ? sirenParsedEntity.actions.map(x => x.name) : [];
+		// todo: sub entities only go one level currently
+		this.entities = sirenParsedEntity.entities ? sirenParsedEntity.entities.map(x => new SirenFacade(x)) : [];
+		this.properties = sirenParsedEntity.properties || {};
+		if (verbose) {
+			this.rawSirenParsedEntity = sirenParsedEntity;
```
AKobets, line 24: Why we need `raw`, could we just have `this.sirenParsedEntity`

state/observable/SirenFacade.js
```diff
@@ -0,0 +1,36 @@
+import { getEntityIDFromSirenEntity } from './ObserverMap.js';
+/**
+ * A class to transform parsed siren entities into easy to use
+ * objects to be reflected and worked with in foundation components.
+ */
+export class SirenFacade {
```
AKobets, line 6: Should it be `SirenEntityFacade`

state/observable/SirenFacade.js
```diff
@@ -0,0 +1,36 @@
+import { getEntityIDFromSirenEntity } from './ObserverMap.js';
```
AKobets, line 1: Future refactoring - `getEntityIDFromSirenEntity` should leave outside of ObserverMap file

state/observable/SirenFacade.js
```diff
@@ -0,0 +1,36 @@
...
+ * A class to transform parsed siren entities into easy to use
+ * objects to be reflected and worked with in foundation components.
+ */
+export class SirenFacade {
+	/**
+	 * Creates the facade from a parsed entity
+	 * @param {Entity} sirenParsedEntity - The raw entity as it has been parsed by SirenParser
+	 * @param {Boolean} verbose - Whether to return more or less information
+	 */
+	constructor(sirenParsedEntity, verbose = false) {
```
AKobets, line 12: why we need verbose and why we do not want to store & provide reference to the parsed entity ?

m6jones: The more generic information, we give to the component the less we will be able to auto document. I have a plan to auto-generate some documentation for components so they stay up to date. 

Giving the entire sub-entity will make it so there are more customizations.

state/observable/SirenFacade.js
m6jones: I like the `raw` as it a word to say, red flag... Explain why you need to go this deep sort and code some ugly stuff.

state/observable/SirenFacade.js
zinabat: Probably

state/observable/SirenFacade.js
zinabat: We decided against having this reference available by default because, to Maya's point, having the reference makes it more likely that developers will misuse it to create very customized (not shareable) components. For this reason, the default is less information.

state/observable/SirenFacade.js
zinabat: Seems like a good candidate to live in the `helper` folder

state/observable/SirenFacade.js
tenbradley: It means if I want to do bad things, I need to specifically turn on the ability to do bad things.

state/observable/SirenFacade.js
tenbradley: Since there is specifically a SirenEntity, I feel that SirenEntityFacade makes me think it's a facade for just SirenEntity. I think SirenFacade is more generic and doesn't seem tied to a specific class.

state/observable/SirenFacade.js
m6jones: This is probably true. I have this weird hate for helper folders. A helper folder is a junk drawer...

state/observable/SirenFacade.js
m6jones: Which will flag me to jump on the PR and really question it.

state/observable/SirenFacade.js
AKobets: lets make it even more ugly `rawRotenSirenParsedEntity`

state/observable/SirenFacade.js
```diff
@@ -0,0 +1,36 @@
...
+		if (!sirenParsedEntity) return;
+
+		this.href = getEntityIDFromSirenEntity(sirenParsedEntity) || undefined;
+		this.class = sirenParsedEntity.class || [];
+		this.links = sirenParsedEntity.links || [];
+		// todo: We should figure out how we want to do actions - this is just a list of names
+		this.actions = sirenParsedEntity.actions ? sirenParsedEntity.actions.map(x => x.name) : [];
+		// todo: sub entities only go one level currently
+		this.entities = sirenParsedEntity.entities ? sirenParsedEntity.entities.map(x => new SirenFacade(x)) : [];
```
AKobets, line 21: should it pass verbose flag to constructed facade for entities ?

zinabat: yes, good catch!

zinabat: I've added the flag and also created a test for this.

AKobets (approved):

Issue #53, opened 2020-12-15
Flakey test: observable with route


zinabat:
https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/master/test/state/HypermediaState.test.js#L116

This test has been failing periodically  with a timeout on the `fetchMock` - we should figure out why.

PR #54, opened 2020-12-15
test: Fixing not flaky test

zinabat:
Fixes issue #53 

Wait for the `fetchStatus` on the relevant states instead of for the `mock` to be called.

### DIFF
--- a/state/HypermediaState.js
+++ b/state/HypermediaState.js
@@ -39,6 +39,18 @@ class HypermediaState extends Fetchable(Object) {
 		});
 	}
 
+	/**
+	 * Hook for this fetch and all children state fetches to complete
+	 * This does not go further than a single nested state currently because state links can be cyclical
+	 * @returns {Promise} Resolves when this fetch and its linked states are all complete
+	 */
+	get allFetchesComplete() {
+		return (async() => {
+			await this.fetchStatus.complete;
+			await Promise.all(this._childStates().map(state => state.fetchStatus.complete));
+		})();
+	}
+
 	createChildState(entityID, token) {
 		token = token === undefined ? this.token.rawToken : token;
 		return stateFactory(entityID, token);
--- a/test/state/HypermediaState.test.js
+++ b/test/state/HypermediaState.test.js
@@ -147,8 +147,11 @@ describe('HypermediaState class', () => {
 
 			await fetch(state);
 
-			await wait(() => mock.called(entityHref));
-			await wait(() => mock.called(linkedHref));
+			await state.fetchStatus.complete;
+			await Promise.all(state._childStates().map(childState => childState.fetchStatus.complete));
+
+			assert.isTrue(mock.called(entityHref));
+			assert.isTrue(mock.called(linkedHref));
 			assert.deepEqual(observer, {
 				mainEntityClass: ['main-entity-class'],
 				linkedEntityProperty: 'linked-entity-name'
--- a/test/state/HypermediaState.test.js
+++ b/test/state/HypermediaState.test.js
@@ -146,9 +146,7 @@ describe('HypermediaState class', () => {
 			state.addObservables(observer, observable);
 
 			await fetch(state);
-
-			await state.fetchStatus.complete;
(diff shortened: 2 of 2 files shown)


test/state/HypermediaState.test.js
```diff
@@ -147,8 +147,11 @@ describe('HypermediaState class', () => {
 
 			await fetch(state);
 
-			await wait(() => mock.called(entityHref));
-			await wait(() => mock.called(linkedHref));
+			await state.fetchStatus.complete;
```
AKobets, line 150: should we have method in HypermediaState that performs the code on lines 150,151 ?

zinabat: hmm, I think it makes sense to have a public method in the state for this.

m6jones: This is coming in the skeleton update. Since it is more complicated.
Shouldn't 
```
await wait(() => mock.called(entityHref));
await wait(() => mock.called(linkedHref));
```
wait for us to fetch both of these?

m6jones: I am getting this error: `WARN: 'Unmatched GET to /[object%20Object]'` I don't want that to happen.

tenbradley: I was also getting this error.

zinabat: odd - I am not? But yes the purpose of the hooks in `fetchStatus` are partly to allow something like `updateComplete`, so devs can fetch for their fetches to complete, especially for testing.

m6jones: `await fetch(state);` === `state.fetchStatus.complete`

zinabat: I can just remove the `await` from the fetch call

zinabat: For some reason if I remove the `await` from the `fetch(state)` line, the test fails :/

state/HypermediaState.js
```diff
@@ -39,6 +39,18 @@ class HypermediaState extends Fetchable(Object) {
 		});
 	}
 
+	/**
+	 * Hook for this fetch and all children state fetches to complete
+	 * This not go further than a single nested state currently because state links can be cyclical
```
tenbradley, line 44: "This not go further" -> "This does not go further"

m6jones:
So the real issue with the test was that fetch wasn't included so we can probably close this PR?

test/state/HypermediaState.test.js
m6jones: This is because `await fetch(state); === await state.fetchStatus.complete` isnt exactly the entire story. `await fetch(state)` returns a promise that has one additional nugget. Which is it also waits for `state.onServerResponse` to be called before resolving the promise. Where `state.fetchStatus.complete` doesn't wait for that step.

zinabat: Aaah, okay

m6jones (approved):
allFetchesComplete sweet!

m6jones:
Before you merge can you run the tests after taking out a fetch import to see if it fails?

state/HypermediaState.js
```diff
@@ -39,6 +39,18 @@ class HypermediaState extends Fetchable(Object) {
...
+	 * @returns {Promise} Resolves when this fetch and its linked states are all complete
+	 */
+	get allFetchesComplete() {
```
AKobets, line 47:
```
async waitForFetchComplete() {
            await this.fetchStatus.complete;
			await Promise.all(this._childStates().map(state => state.fetchStatus.complete));
}
```

zinabat:
> fetch
Removing fetch import from `HypermediaState` results in the following:

Screen Shot 2020-12-17 at 11 48 19 AM

state/HypermediaState.js
zinabat: I prefer the current way because we can reference it as a property similar to `updateComplete` (Maya is working on something similar for skeletons) - I did originally have it as a method, though.

PR #57, opened 2020-12-18
Fix: entityID has no method to lower its case

m6jones:


state/StateStore.js
```diff
@@ -16,6 +16,10 @@ export class StateStore {
 	}
 
 	get(entityID, token) {
+		if (!entityID || !token?.toString()) {
```
zinabat, line 19: What does this second check `!token?.toString()` do? Like what are the cases it will case?

m6jones: It looks cool. It checks if we have a token and it isn't empty.

zinabat (approved):

PR #58, opened 2020-12-22
feat: Implement SirenFacade for relevant Observable types

zinabat:
## Context
[US123074](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F461413871352&fdp=true?fdp=true)

The second and final portion of the `SirenFacade` story connects the new `SirenFacade` class into the following `Observable`s:

- `SirenSubEntity`: now reflects the `SirenFacade` instead of the parsed entity
- `SirenSubEntities`: reflects an Array of `SirenFacade` objects
- `SirenSummonAction`: reflects the same action, but returns a `SirenFacade` on `summon()`
- Added the `rel` property to the `SirenFacade`

## Quality
- Existing unit tests

### DIFF
--- a/state/observable/sirenObservableFactory.js
+++ b/state/observable/sirenObservableFactory.js
@@ -33,14 +33,13 @@ const observableClasses = Object.freeze({
  *
  * @param {*} param0
  */
-function definedProperty({ observable: type, prime, rel: id, route, token, state, verbose }) {
+function definedProperty({ observable: type, prime, rel: id, route, token, state }) {
 	return {
 		id,
 		route,
 		token: (prime || route) ? token : undefined,
 		type,
-		state,
-		verbose
+		state
 	};
 }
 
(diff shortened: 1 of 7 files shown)


state/observable/sirenObservableFactory.js
```diff
@@ -29,13 +29,18 @@ const observableClasses = Object.freeze({
...
+ */
+function definedProperty({ observable: type, prime, rel: id, route, token, state, verbose }) {
 	return {
 		id,
 		route,
 		token: (prime || route) ? token : undefined,
 		type,
-		state
+		state,
+		verbose
```
m6jones, line 43: Since not all of the observer types need verbose. Should these be defined in the static method?

zinabat: Which method? I only put it here because `prime` and `route` are defined here, but is there a better place?

zinabat: Moved where this is defined.

m6jones (approved):

PR #66, opened 2021-01-06
feat: Skeleton!

m6jones:
image

https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/56

### DIFF
--- a/render/componentFactory.js
+++ b/render/componentFactory.js
@@ -39,6 +39,10 @@ class ComponentStore {
 		});
 	}
 
+	skeletonTag() {
+		return this._componentStore.get(defaultType);
+	}
+
 	_getComponentTag(type, otherTypes, map, depth = 0) {
 		map = map ? map : this._componentStore;
 		if (!map.has(type)) {
--- a/framework/lit/HypermediaResult.js
+++ b/framework/lit/HypermediaResult.js
@@ -198,14 +198,18 @@ export class HypermediaResult extends TemplateResult {
 		};
 		const resources = { classes: [] };
 		const components = componentStoreFactory(pseudoTag);
-		if (!href || !token) return html`loading`;
+		if (!href || !token) return this._skeletonRender(components);
 		const statePromise = stateFactory(href, token);
 		const fetchedResults  = statePromise.then(async state => {
 			state.addObservables(resources, observable);
 			await fetch(state);
 			return state;
 		}).then(state => this.render(state, components, resources, pseudoTag, strings, values));
-		return html`${until(fetchedResults, html`loading`)}`;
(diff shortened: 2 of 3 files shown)


render/componentFactory.js
```diff
@@ -39,6 +39,10 @@ class ComponentStore {
 		});
 	}
 
+	skeletonTag() {
+		return this._componentStore.get(defaultType);
```
m6jones, line 43: Right now it gets the default tag. Seems to be pretty good to just do that. Should be easy to change if we have too.

tenbradley (commented):
Overall, this looks good. My only concern is what if there is nothing to load? For instance, looking at an activity without an image or a title results in it loading indefinitely.

But overall, it seems to work. Huzzah.

state/HypermediaState.js
```diff
@@ -38,22 +40,38 @@ class HypermediaState extends Fetchable(Object) {
...
 	 */
 	async allFetchesComplete(skipTheseStates) {
+		await this._waitForFirstFetch;
+		await this.fetchStatus.complete;
 		skipTheseStates = Array.isArray(skipTheseStates) ? skipTheseStates : [skipTheseStates];
 		const skipTheseStatesOnNextStep = [ this, ...skipTheseStates ];
-		await this.fetchStatus.complete;
+		const parentPromises = [];
+		this._parents.forEach((_, parent) => {
+			if (!skipTheseStates.includes(parent)) {
```
zinabat, line 58: This prevents circular referencing?

test/state/HypermediaState.test.js
```diff
@@ -677,13 +674,42 @@ describe('HypermediaState class', () => {
...
 			const maxDelay = delayFetchAmountMs * depthOfResponses;
+			expect(observer.parentAnswer).to.equal(42);
+			expect(observer.subEntityFirstAnswer).to.equal(43);
+			expect(observer.subEntityFirstSubAnswer).to.equal(44);
+			expect(observer.subEntitySecondAnswer).to.equal(45);
+			expect(observer.primeLinkAnswer).to.equal(46);
 			expect(timeAfterFetch - timeBeforeFetch).to.be.above(maxDelay);
+		});
+
+		it('the child state, should know when the parent and it\'s children have loaded', async() => {
```
zinabat, line 685: Do we have a test for circular references? Would it be worth adding one if not?

zinabat (approved):
Just one thing to check but tested and it looks good.

state/HypermediaState.js
m6jones: Yup :)

test/state/HypermediaState.test.js
m6jones: I do! There a circular reference in my test data.

PR #67, opened 2021-01-07
fix: Remove await from the fetch

zinabat:
The `await` breaks the localizeMixin from core.


tenbradley (approved):
Grr localizeMixin causing headaches.

PR #68, opened 2021-01-08
feat: Added support for array fields

zinabat:
## Context
[US122856](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F459739288064&fdp=true?fdp=true)

Engine portion of the story. The new `bulk-update-collection` action takes fields in the form of:
```
inputIds: 1
inputIds: 2
```
Previously, the engine did not support fields with the same name. I've remedied this in this PR. Now, in the component, we do this with an action like so:
```js
this.someAction.commit({
    inputIds: [1, 2]
});
```
The engine will automagically unpack it into multiple formData entries.

## Quality
- New very basic unit tests added for `SirenAction`. None existed before. We should add more.
- Basic integration test for `actions` added
- Cleaned up existing integration tests so they no longer use `aTimeout`

### DIFF
--- a/test/example-components/component-integration.test.js
+++ b/test/example-components/component-integration.test.js
@@ -1,101 +1,107 @@
 import './summon-action-components.js';
 import './subentities-components.js';
-import { aTimeout, expect, fixture, waitUntil } from '@open-wc/testing';
+import './action-components.js';
+import { expect, fixture, waitUntil } from '@open-wc/testing';
 import { default as fetchMock } from 'fetch-mock/esm/client.js';
 import { html } from 'lit-html';
 
 describe('Component integration', () => {
-	it('receives an entity when we summon one', async() => {
-		const selfHref = 'https://summon-action/entity';
-		const actionHref = 'https://summon-action/do';
-		const entity = {
-			actions: [{ href: actionHref, method: 'POST', name: 'example-summon' }],
-			links: [{ rel: ['self'], href: selfHref }]
-		};
-		const summonedEntity = {
-			properties: [{ someProperty: 'foobar' }]
-		};
-		const mock = fetchMock
-			.mock(selfHref, JSON.stringify(entity), {
-				delay: 50, // fake a slow network
-			})
-			.post(actionHref, JSON.stringify(summonedEntity), {
-				delay: 100, // fake a slow network
-			});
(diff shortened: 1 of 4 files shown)


test/observable/sirenAction.test.js
```diff
@@ -1,2 +1,49 @@
...
+			action._fields = { 'someField': 'someValue' };
+			action._rawSirenAction = { type: ['json'], method: 'POST' };
+			action.setBodyFromInput({ 'anotherField': 'anotherValue' });
+			expect(action._body, 'body is a string').to.be.a('string');
+			expect(JSON.parse(action._body), 'body equals the json').to.deep.equal({
+				'anotherField': 'anotherValue', 'someField': 'someValue'
+			});
+		});
+
+		it('does not set the body for GET and HEAD requests', () => {
```
AKobets, line 19: test actually does not cover HEAD request.

test/example-components/component-integration.test.js
```diff
@@ -1,99 +1,105 @@
...
-
-		await waitUntil(() => mock.called(actionHref));
-
-		await aTimeout(400);
-		expect(element._hasAction('nestedSummon'), 'has the action').to.be.true;
+		it('gets an summon action routed through a summon action', async() => {
+			const selfHref = 'https://summon-action/entity-3';
+			const actionHref = 'https://summon-action/do-3';
+			const entity = {
+				actions: [{ href: actionHref, method: 'POST', name: 'example-summon' }],
```
AKobets, line 85: Do we have any test that asserts that actionHref is called correctly: expected Method, Content Type , action fields/values ?

test/observable/sirenAction.test.js
zinabat: oops, will fix

test/example-components/component-integration.test.js
zinabat: So I couldn't figure out how to get the action fields/values but if you have tips on how to spy on the `fetch-mock` call please tell me. I can definitely add assertions for method and content type though.

m6jones: The test in fetch would check if we are calling it. So testing if the actionHref is set in SirenAction class is good enough for this release.

m6jones (approved):

test/observable/sirenAction.test.js
zinabat: fixed

PR #77, opened 2021-01-14
Expose the error to the component

zinabat:
Exposes `_error` to the component, which is just a Boolean that tells us whether there is an error or not. We can change this to be more informative later.


m6jones (approved):

zinabat:
Closing this so that improvements can be made to this idea _with tests_

PR #80, opened 2021-02-03
fix: polyfill missing matchAll in legacy edge

stevenao:


framework/lit/HypermediaResult.js
```diff
@@ -4,6 +4,17 @@ import { observableTypes, stateFactory } from '../../state/HypermediaState.js';
 import { fetch } from '../../state/fetch.js';
 import { until } from 'lit-html/directives/until.js';
 
+if (!String.prototype.matchAll) {
+	String.prototype.matchAll = function(rx) {
```
AKobets, line 8: unit test needed

zinabat (changes requested):
We need to use `match` instead of adding a polyfill like @dlockhart suggested - I can do that today unless you want to.

PR #81, opened 2021-02-05
fix: Use while exec instead of matchAll

zinabat:
## Context
[US124640](https://rally1.rallydev.com/#/357252966636d/dashboard?detail=%2Fuserstory%2F492652604048&fdp=true?fdp=true)
https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/80

Changes the `matchAll` function within the `HypermediaResults` class to use a `while` loop with the `exec` function to perform regex processing because `matchAll` is not supported in Legacy Edge. This solution was chosen over a polyfill.

**Note:** `exec` was chosen over the `match` function because the global flag is required. `match` will ignore capture groups whenever the global flag is set, where exec does not. See also: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#description

New integration tests added to test the functionality of the custom `html` function, the `customHypermediaElement` function, and the processing performed by `HypermediaResult`.

## Quality
- New integration tests added for previously uncovered code
- There is a new Promise object exposed in `HypermediaResult` which is strictly for testing. This was to avoid needing to mock out the functionality of the `until` directive. It can likely be removed once we have an alternative to `fixture` from `open-wc/testing` - currently our `HypermediaResult` does not work with `fixture`

### DIFF
--- a/framework/lit/HypermediaResult.js
+++ b/framework/lit/HypermediaResult.js
@@ -75,8 +75,7 @@ export class HypermediaResult extends TemplateResult {
 		}
 
 		// return all of this information as a TemplateResult to be stored as a value for a different Result
-		const result = new TemplateResult(mainStrings, mainValues, 'html', defaultTemplateProcessor);
-		return result;
+		return new TemplateResult(mainStrings, mainValues, 'html', defaultTemplateProcessor);
 	}
 
 	/*
--- a/framework/lit/HypermediaResult.js
+++ b/framework/lit/HypermediaResult.js
@@ -75,8 +75,8 @@ export class HypermediaResult extends TemplateResult {
 		}
 
 		// return all of this information as a TemplateResult to be stored as a value for a different Result
-
-		return new TemplateResult(mainStrings, mainValues, 'html', defaultTemplateProcessor);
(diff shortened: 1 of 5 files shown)


m6jones (approved):

framework/lit/HypermediaResult.js
```diff
@@ -75,19 +75,23 @@ export class HypermediaResult extends TemplateResult {
...
 	/*
 	*  Determines the href and token associated with the current element
 	*/
-
 	getHrefToken(strings, values) {
 		let href, token;
 		strings = [...strings];
 		strings.forEach((string, index) => {
-			const output = [...string.matchAll(/(token=)|(href=)/g)].pop();
+			const matches = [];
```
AKobets, line 88: would it be simpler to use str.match(regex) ?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

m6jones: I believe because we are using the global regex, we can't use match.

zinabat: correct. `match` ignores capture groups when the global flag is set

PR #82, opened 2021-02-09
fix: Update Deep copy to handle cycles and added tests.

m6jones:
If there was a cycle in objects we were deep copying we would receive a stackover flow. So I fixed it and added tests to deep copy and how we expect things to work.

### DIFF
--- a/helper/deepCopy.js
+++ b/helper/deepCopy.js
@@ -1,4 +1,4 @@
-export function deepCopy(inObject) {
+export function deepCopy(inObject, copiedObjects = null) {
 	if (typeof inObject !== 'object' || inObject === null) {
 		return inObject; // Return the value if inObject is not an object
 	}
@@ -6,12 +6,19 @@ export function deepCopy(inObject) {
 	// Create an array or object to hold the values
 	const outObject = Array.isArray(inObject) ? [] : {};
 
+	// What if there is a cycle?
+	copiedObjects = copiedObjects === null ? new WeakMap() : copiedObjects;
+	if (copiedObjects.has(inObject)) {
+		return copiedObjects.get(inObject);
+	}
+	copiedObjects.set(inObject, outObject);
+
 	let value;
 	for (const key in inObject) {
 		value = inObject[key];
 
 		// Recursively (deep) copy for nested objects, including arrays
-		outObject[key] = deepCopy(value);
+		outObject[key] = deepCopy(value, copiedObjects);
 	}
 
 	return outObject;
--- a/test/example-components/component-integration.test.js
+++ b/test/example-components/component-integration.test.js
@@ -248,7 +248,9 @@ describe('Component integration', () => {
 				await elements[0].updateItems(newItems);
 
 				elements.forEach(element => {
+					// copy not a ref
 					expect(element.items).to.deep.equal(newItems);
+					expect(element.items).to.not.equal(newItems);
 				});
 			});
 		});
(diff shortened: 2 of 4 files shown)


helper/deepCopy.js
```diff
@@ -1,17 +1,24 @@
...
 	// Create an array or object to hold the values
 	const outObject = Array.isArray(inObject) ? [] : {};
 
+	// What if there is a cycle?
+	copiedObjects = copiedObjects === null ? new WeakMap() : copiedObjects;
+	if (copiedObjects.has(inObject)) {
+		return copiedObjects.get(inObject);
+	}
+	copiedObjects.set(inObject, outObject);
+
```
m6jones, line 15: Keep track of which objects we seen in a weak map. Weak map to make sure we don't have a memory leak and since we don't need to iterate through them.

test/example-components/component-integration.test.js
```diff
@@ -128,6 +130,38 @@ describe('Component integration', () => {
...
+			}
+			await waitUntil(() => elements[0]._loaded === true);
+			elements.forEach(element => {
+				expect(mock.called(selfHref)).to.be.true;
+				expect(element.items).to.exist;
+				expect(element.items).to.be.an('array');
+				expect(element.items).to.have.lengthOf(5);
+				let uniqueId = null;
+				element.items.forEach((item, key) => {
+					uniqueId = uniqueId === null ? item.properties.itemNumber : uniqueId;
```
m6jones, line 159: I get a random number and add it to the index of itemNumber. So the first one has index 0 and so `0+uniqueId = uniqueId` thus I can use this to test the rest of the numbers to make sure it works out.

test/helper/deepCopy.test.js
```diff
@@ -0,0 +1,130 @@
...
+		undefined: 'not undefined',
+		null: 'not null',
+		function: () => 'new method',
+		object: {
+			prop1: 3,
+			prop2: 4,
+		},
+		array: [ 2, 3, 4, 5]
+	};
+	const changeDefaultObject = (object) => {
```
m6jones, line 18: This method will change the object to look like `changedObject`

test/helper/deepCopy.test.js
```diff
@@ -0,0 +1,130 @@
...
+		object.string = 'Fluffy';
+		object.undefined = 'not undefined';
+		object.null = 'not null';
+		object.function = () => 'new method';
+		object.object.prop1 = 3;
+		object.object.prop2 = 4;
+		object.array.push(5);
+		object.array.shift();
+	};
+	const compareObjectsValues = (a, b) => {
```
m6jones, line 29: I want to be verbose on which part of the object didn't match if a test fails.

helper/deepCopy.js
zinabat: yesssss nice

zinabat (commented):
Can you write a test that updates the state from one of the observing components? I noticed you wrote a component with a method `updateItems` but then didn't use it

m6jones:
> Can you write a test that updates the state from one of the observing components? I noticed you wrote a component with a method `updateItems` but then didn't use it

https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/82/files#diff-d492e147520be29e7b7d0b5711d8ad22d6d86a4f8d14c2a7ea646288abe11143R133

zinabat (approved):

PR #85, opened 2021-03-06
fix: Adding check to prevent weird caching issue

zinabat:
## Context
https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/128

I wanted to write a test for this, but couldn't reproduce it in the engine. This seems to only be a problem for Safari.

### DIFF
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -60,13 +60,13 @@ export class SirenSubEntities extends Observable {
 			const entityID = getEntityIDFromSirenEntity(sirenSubEntity);
 			let subEntity;
 			// If we already set it up why do it again?
-			if (this.entityMap.has(entityID)) {
+			if (entityID && this.entityMap.has(entityID)) {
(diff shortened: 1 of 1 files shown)


m6jones (approved):

state/observable/SirenSubEntities.js
```diff
@@ -60,13 +60,13 @@ export class SirenSubEntities extends Observable {
 			const entityID = getEntityIDFromSirenEntity(sirenSubEntity);
 			let subEntity;
 			// If we already set it up why do it again?
-			if (this.entityMap.has(entityID)) {
+			if (entityID && this.entityMap.has(entityID)) {
 				subEntity = this.entityMap.get(entityID);
 			} else {
 				subEntity = new SirenSubEntity({ id: this.rel, token: this._token, verbose: this._verbose, state: this._state });
```
tenbradley, line 66: If we enter this case with entityID being 'false', should we be storing it in the map with that ID?

PR #90, opened 2021-03-30
feat: A couple of small but neat updates

m6jones:
Fixed in this PR:
- Deep Copy didn't work for dates. Now it does.

New Features:
- Before we make a state, we can wait for other attributes to be given a value. This is useful when you need to setup observables that require non-static properties.
- You can now pass custom properties at addObservable stage. This is useful if you have a more complex custom observable and the output might change based on some of these options.

### DIFF
--- a/framework/lit/HypermediaStateMixin.js
+++ b/framework/lit/HypermediaStateMixin.js
@@ -28,6 +28,7 @@ export const HypermediaStateMixin = superclass => class extends superclass {
 		super();
 		this._loaded = false;
 		this.__observables = this.constructor.properties;
+		this.__waitForAttributes = {};
 	}
 
 	connectedCallback() {
@@ -43,11 +44,30 @@ export const HypermediaStateMixin = superclass => class extends superclass {
 	}
 
 	updated(changedProperties) {
-		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token && this.href !== 'undefined') {
+		if (this.__shouldUpdateState(changedProperties)) {
 			dispose(this._state, this);
 			this._makeState();
 		}
 		super.updated(changedProperties);
+	}
+
+	waitForProperty(attribute, valuesThatAreFalsy = []) {
+		if (!this.__waitForAttributes[attribute]) {
+			this.__waitForAttributes[attribute] = [];
+		}
+		this.__waitForAttributes[attribute] = [...this.__waitForAttributes[attribute], ...valuesThatAreFalsy];
+	}
+
+	__shouldUpdateState(changedProperties) {
+		const propertiesNeedToHaveValues = { ...this.__waitForAttributes, href: ['undefined'], token: [] };
+		const onePropertyIsChanging = Object.keys(propertiesNeedToHaveValues).some(property => changedProperties.has(property));
+		if (!onePropertyIsChanging) return false;
+
+		const value = !Object.keys(propertiesNeedToHaveValues).some(property => {
+			const isAFalsyValue = propertiesNeedToHaveValues[property].some(value => this[property] === value);
+			return !this[property] || isAFalsyValue;
+		});
+		return value;
 	}
 
 	_hasAction(action) {
--- a/framework/lit/HypermediaStateMixin.js
+++ b/framework/lit/HypermediaStateMixin.js
@@ -51,7 +51,7 @@ export const HypermediaStateMixin = superclass => class extends superclass {
 		super.updated(changedProperties);
 	}
 
-	waitForProperty(property, valuesThatAreFalsy = []) {
+	requiredPropertyForState(property, valuesThatAreFalsy = []) {
 		if (!this.__waitForAttributes[property]) {
 			this.__waitForAttributes[property] = [];
 		}
--- a/framework/lit/HypermediaStateMixin.js
+++ b/framework/lit/HypermediaStateMixin.js
@@ -51,23 +51,22 @@ export const HypermediaStateMixin = superclass => class extends superclass {
 		super.updated(changedProperties);
 	}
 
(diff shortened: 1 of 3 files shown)


framework/lit/HypermediaStateMixin.js
```diff
@@ -43,13 +44,32 @@ export const HypermediaStateMixin = superclass => class extends superclass {
...
 	updated(changedProperties) {
-		if ((changedProperties.has('href') || changedProperties.has('token')) && this.href && this.token && this.href !== 'undefined') {
+		if (this.__shouldUpdateState(changedProperties)) {
 			dispose(this._state, this);
 			this._makeState();
 		}
 		super.updated(changedProperties);
 	}
 
+	waitForProperty(attribute, valuesThatAreFalsy = []) {
```
m6jones, line 54: This is the method that allows you to wait for an attribute to be set.
The name might not be great so taking suggestions.

The added parameter, `valuesThatAreFalsy` is an array of other falsy values that should be considered as the attribute is empty. For exampe, `href` being `"undefined"` is falsy for that attribute.

state/HypermediaState.js
```diff
@@ -36,7 +36,7 @@ class HypermediaState extends Fetchable(Object) {
 			const definedProperty = sirenObserverDefinedProperty(propertyInfo, this);
 			if (!definedProperty) return;
 			const sirenObservable = this._getSirenObservable(definedProperty);
-			sirenObservable.addObserver(observer, name, { route: definedProperty.route ? { [name]: definedProperty.route } : undefined, method: observables[name].method });
+			sirenObservable.addObserver(observer, name, { ...observables[name], route: definedProperty.route ? { [name]: definedProperty.route } : undefined });
```
m6jones, line 39: Adding all the properties and then rewriting route. This is probably the scariest change but all the test past so yay!

helper/deepCopy.js
```diff
@@ -3,6 +3,10 @@ export function deepCopy(inObject, copiedObjects = null) {
 		return inObject; // Return the value if inObject is not an object
 	}
 
+	if (inObject instanceof Date) {
+		return new Date(inObject.getTime());
```
m6jones, line 7: If the object we are deep copy is a date. Lets just copy it this way instead so it will work :D

framework/lit/HypermediaStateMixin.js
zinabat: The name makes this feel like it returns a promise, but I see that it does not. What are these things exactly? Required attributes? What's the difference between an attribute and a property in this case - I see both names are used?

tenbradley: nit: would it make sense for the parameter attribute to be named property if the function name is not changed? I think the parameter should match the function name?

helper/deepCopy.js
tenbradley: Is there a reason that this deep copy has to be different than others? This if scares me that we might end up having a lot of these types of conditions in the future.

framework/lit/HypermediaStateMixin.js
```diff
@@ -43,13 +44,32 @@ export const HypermediaStateMixin = superclass => class extends superclass {
...
+	__shouldUpdateState(changedProperties) {
+		const propertiesNeedToHaveValues = { ...this.__waitForAttributes, href: ['undefined'], token: [] };
+		const onePropertyIsChanging = Object.keys(propertiesNeedToHaveValues).some(property => changedProperties.has(property));
+		if (!onePropertyIsChanging) return false;
+
+		const value = !Object.keys(propertiesNeedToHaveValues).some(property => {
+			const isAFalsyValue = propertiesNeedToHaveValues[property].some(value => this[property] === value);
+			return !this[property] || isAFalsyValue;
+		});
+		return value;
```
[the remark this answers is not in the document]
tenbradley: Returns true if all properties specified have a non-Falsy value, false otherwise is what I read.

helper/deepCopy.js
m6jones: I think Date and maybe some other prebuilt ones might not be happy with our deep copy. The error was super obvious so it should be easy to find in the future.

framework/lit/HypermediaStateMixin.js
m6jones: Updated the name to `requiredPropertyForState` thoughts?

framework/lit/HypermediaStateMixin.js
```diff
@@ -43,13 +44,32 @@ export const HypermediaStateMixin = superclass => class extends superclass {
...
 
+	waitForProperty(attribute, valuesThatAreFalsy = []) {
+		if (!this.__waitForAttributes[attribute]) {
+			this.__waitForAttributes[attribute] = [];
+		}
+		this.__waitForAttributes[attribute] = [...this.__waitForAttributes[attribute], ...valuesThatAreFalsy];
+	}
+
+	__shouldUpdateState(changedProperties) {
+		const propertiesNeedToHaveValues = { ...this.__waitForAttributes, href: ['undefined'], token: [] };
```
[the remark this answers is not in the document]
m6jones: updated, much better

framework/lit/HypermediaStateMixin.js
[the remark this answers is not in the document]
m6jones: So `some` breaks out and return true on the first truth value.

I wanted to make an `and` so I would only return true if one of the values is falsy. 
So this basically loops through the list until it finds a value that is falsy and returns true.

framework/lit/HypermediaStateMixin.js
zinabat: I'm fine with that

tenbradley (approved):

PR #92, opened 2021-04-05
fix: we haven't been paying attention to subentities relationships. 

m6jones:
This is a defect:https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F518693719340
It shows up using this branch: https://github.com/Brightspace/quiz-components/tree/US124788/state_push_fix

The issue is we want to call push on the parent and for it to look at all of it children. But right now the children are subentities and well, that wasn't supported / didn't work. This is to fix that problem.

I've been testing this since last Friday on my local instance and has been working!

### DIFF
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -46,6 +46,12 @@ export class SirenSubEntities extends Observable {
 		return this._rel;
 	}
 
+	get routedState() {
+		const routedStates = [];
+		this.entityMap.forEach(subEntity => subEntity.routedState && routedStates.push(subEntity.routedState));
+		return routedStates;
+	}
+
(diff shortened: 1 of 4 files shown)


state/observable/SirenSubEntities.js
```diff
@@ -45,6 +45,12 @@ export class SirenSubEntities extends Observable {
 		return this._rel;
 	}
 
+	get routedState() {
```
tenbradley, line 48: I think this should be routedStates to convey there could be more than one.

m6jones: It has to be consistent with the others. I am thinking of overhauling the building of the graph so I just kept it to be routedState like in the mixin.

tenbradley (approved):

PR #94, opened 2021-04-07
fix: Entities were added out of order.

m6jones:
fix for, "a bug where Quiz Sections appear to be getting out of order when they are created using the Save&Copy or Save&New FACE Quiz workflow. Looking at the collections api response, you can see that the incoming data is in ascending order by id [220, 221, 222, 223]. But when I log this.items in d2l-activity-collection-editor-quiz, the order is [223, 220, 221, 222]."

### DIFF
--- a/state/observable/SirenSubEntities.js
+++ b/state/observable/SirenSubEntities.js
@@ -1,5 +1,6 @@
 import { getEntityIDFromSirenEntity } from './ObserverMap.js';
 import { Observable } from './Observable.js';
+import { SirenFacade } from './SirenFacade.js';
 import { SirenSubEntity } from './SirenSubEntity.js';
 
 /**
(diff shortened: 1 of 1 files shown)


state/observable/SirenSubEntities.js
```diff
@@ -67,7 +69,6 @@ export class SirenSubEntities extends Observable {
 				await subEntity.setSubEntity(sirenSubEntity);
```
m6jones, line 69: This await was the problem!

tenbradley: If this await was the problem, how is it solved?

m6jones: I moved adding to the list above the await. So it happens in sync?

state/observable/SirenSubEntities.js
```diff
@@ -58,6 +59,7 @@ export class SirenSubEntities extends Observable {
 		// and will change based on the individual item update.
 		const promises = subEntities.map(async(sirenSubEntity) => {
```
m6jones, line 60: This is the start of the loop of an async method. So each method goes and gets called and everything is done up until the await.

state/observable/SirenSubEntities.js
```diff
@@ -58,6 +59,7 @@ export class SirenSubEntities extends Observable {
...
 			const entityID = getEntityIDFromSirenEntity(sirenSubEntity);
+			sirenFacades.push(new SirenFacade(sirenSubEntity, this._verbose));
```
m6jones, line 62: I moved adding to the list above the await so it should happen in order and not dependent on when the await returns.

tenbradley (approved):

PR #97, opened 2021-04-26
fix: up actions fields.

m6jones:
Basically, we were not handling action GET fields properly. This fixes that issue. We now sub in defaults correctly and allow one to modify them. SO yay!

### DIFF
--- a/state/Fetchable.js
+++ b/state/Fetchable.js
@@ -118,7 +118,7 @@ export const Fetchable = superclass => class extends superclass {
 	 * @returns {String} href that identifies the fetchable
 	 */
 	get href() {
-		return this._setupHrefWithQueryParams();
+		return this._setupHrefWithQueryParams(this._paramsObj);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


state/Fetchable.js
```diff
@@ -168,12 +168,12 @@ export const Fetchable = superclass => class extends superclass {
 	 * E.g. { key: 'input-for-key' } or { key: ['input1', 'input2']}
 	 * @returns {String} A URL containing the query string
 	 */
-	_setupHrefWithQueryParams() {
-		if (!this._paramsObj || this._paramsObj.length === 0) return this._href;
+	_setupHrefWithQueryParams(paramsObj) {
```
tenbradley, line 171: This now never uses the this._paramsObj. Could there be a situation were we want to use it?

m6jones: It gets passed in now so that is why.

tenbradley (approved):

PR #102, opened 2021-05-26
feat: Add bypassCache and tests to summon action

BhamelinD2L:
Adds a `bypassCache` boolean argument to the `summonAction.summon()` function. 
Also adds two tests to confirm that the cached/bypassed calls are caching and bypassing as expected.

I also tested this manually by linking up with the summon action for the [Enrollment Rule match count mixin](https://github.com/BrightspaceHypermediaComponents/foundation-components/blob/5b1a0e01759305ff8937fe34455a265c40fff430/features/discover/mixins/match-count-mixin.js#L28) and reviewing the behavior when the argument is set/unset. This PR is needed to prevent it from caching the first returned count value.

### DIFF
--- a/state/observable/SirenSummonAction.js
+++ b/state/observable/SirenSummonAction.js
@@ -81,11 +81,15 @@ export class SirenSummonAction extends Routable(SirenAction) {
 	_updateAction() {
 		this.action = {
 			has: true,
-			summon: async(observables) => {
+			summon: async(observables, bypassCache) => {
 				this._prepareAction(observables);
 				this._readyToSend = true;
-				// todo: we should be able to just return what we get back from fetch, but it is returning the json
-				await fetch(this);
+
+				const response = await fetch(this);
+				if (bypassCache) {
+					return response;
+				}
+
 				return new SirenFacade(this.routedState._entity, this._verbose);
 			}
 		};
--- a/state/observable/SirenSummonAction.js
+++ b/state/observable/SirenSummonAction.js
@@ -85,11 +85,10 @@ export class SirenSummonAction extends Routable(SirenAction) {
 				this._prepareAction(observables);
 				this._readyToSend = true;
 
-				const response = await fetch(this);
-				if (bypassCache) {
-					return response;
+				if (bypassCache && this.routedState) {
+					this.routedState.byPassCache();
 				}
-
+				await fetch(this);
 				return new SirenFacade(this.routedState._entity, this._verbose);
 			}
 		};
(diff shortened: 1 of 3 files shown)


state/observable/SirenSummonAction.js
```diff
@@ -81,11 +81,15 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
 		this.action = {
 			has: true,
-			summon: async(observables) => {
+			summon: async(observables, bypassCache) => {
 				this._prepareAction(observables);
 				this._readyToSend = true;
-				// todo: we should be able to just return what we get back from fetch, but it is returning the json
-				await fetch(this);
+
+				const response = await fetch(this);
```
zinabat, line 88:
```suggestion
				const response = await fetch(this, bypassCache);
```
I'm fairly confident that the routedState is set by the call to `fetch`. I'm surprised that the tests work 🤔 https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/master/state/fetch.js#L16

zinabat: You should be able to just pass the flag to fetch and leave everything else as is

BhamelinD2L: I see what you're saying, yeah. I've passed `true` to `fetch` like you noted above, but it actually still returns the cached result. I will figure out what's going on.

state/observable/SirenSummonAction.js
```diff
@@ -81,10 +81,13 @@ export class SirenSummonAction extends Routable(SirenAction) {
 	_updateAction() {
 		this.action = {
 			has: true,
-			summon: async(observables) => {
+			summon: async(observables, bypassCache) => {
 				this._prepareAction(observables);
 				this._readyToSend = true;
-				// todo: we should be able to just return what we get back from fetch, but it is returning the json
+
```
BhamelinD2L, line 87: I am getting the response from fetch rather than json, so I'm removing this comment unless there's an objection

state/observable/SirenSummonAction.js
```diff
@@ -81,10 +81,13 @@ export class SirenSummonAction extends Routable(SirenAction) {
...
+				if (bypassCache && this.routedState) {
+					this.routedState.byPassCache();
+				}
 				await fetch(this);
```
BhamelinD2L, line 91: I need to call `bypassCache()` before fetching, as waiting until after the fetch is too late. 
This means I have to confirm `routedState` exists beforehand as well, as it won't exist the first time this runs.

state/observable/SirenSummonAction.js
BhamelinD2L: Follow up: Investigated and discussed this with Maya and Zina, and edited the PR based on our  conclusions

m6jones (approved):

PR #104, opened 2021-06-14
feat: Fetch precache from actions

m6jones:
Will process the pre cache header links


mbardakovD2L (approved):
LGTM - not sure what doors this opens for us but it looks backwards-compatible so that's good

PR #106, opened 2021-06-24
fix: Add request map

zinabat:
## Context
[DE43763](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fdefect%2F600611490877&fdp=true?fdp=true)

Fixes a regression in Learning Paths caused by the server sending a cached response. There is probably a cleaner way to do this, but this was the easiest thing to read I could think of.

**Key Change**:
- Adds a map of `href` keys to ensure we _always_ set no-cache headers on first request

## Quality
**Manual test for defect fix:**
- Pull this branch and npm link it to BSI
- Start a local dev BSI
- Log into the `spacelocalbsi` Sitelord instance
- Change a learning path's name or description
- Visit the learning path's edit page again and note that the name and description are updated

No unit test was added because this is an LMS integration issue.

### DIFF
--- a/state/fetch.js
+++ b/state/fetch.js
@@ -3,6 +3,8 @@ import { myLoadingPromise } from './loader.js';
 
 const d2lfetch = window.d2lfetch;
 
+const serverRequests = new Map();
+
 /**
  * This method sends request to the server. It returns the results to fetchable.onServerResponse and the returned promise.
  * This method also handles cachePrimer header from the server with links with the rel, https://api.brightspace.com/rels/cache-primer.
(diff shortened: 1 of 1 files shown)


state/fetch.js
```diff
@@ -55,13 +57,15 @@ async function performServerFetch(fetchable, bypassCache) {
 	const fetch = !fetchable.token.cookie ? d2lfetch : d2lfetch.removeTemp('auth');
 
 	const headers = fetchable.headers;
-	if (bypassCache) {
+	// force a bypass cache if we haven't gotten the href yet on this page
+	if (bypassCache || !serverRequests.has(fetchable.href)) {
```
m6jones, line 61: This may slow down things since the cache is suppose to help pages that are loading similar data.

zinabat: possible, but I still think this is preferable to forcing consumers of the engine to specify `no-cache` headers on first load

zinabat: Or maybe not, I'm trying to think of the easiest way we could force users to specify to set `no-cache` headers on first load. Perhaps yet another property attribute?

m6jones: I think maybe on actions that return an entity we force it to bust the cache? I dont understand how the backend doesn't do this..

PR #107, opened 2021-07-05
fix: cache primer not being refreshed

stevenao:


state/fetch.js
```diff
@@ -65,7 +65,9 @@ async function performServerFetch(fetchable, bypassCache) {
 	if (!response.ok) {
 		throw response.status;
 	}
-	await fetchable.handleCachePriming(cachePrimingList(response));
+	setTimeout(() => {
+		fetchable.handleCachePriming(cachePrimingList(response));
+	}, 0);
```
zinabat, line 70: I'm surprised this works at all - don't we still need to wait for the promise to resolve? It breaks a test. :(

stevenao: @zinabat It actually works quite well in my testing. The idea is to queue it after the `_state.push`.  However, I am not sure about this LoC myself. It is more to illustrate the problem why "link header" state is not refreshing correctly. You mentioned early that you have another approach in mind. Maybe that is something I can test it here if you can provide the code.

state/HypermediaState.js
```diff
@@ -106,7 +106,7 @@ class HypermediaState extends Fetchable(Object) {
 		token = token !== null ? token : this.token;
 		return Promise.all(links.map(async(link) => {
 			const state = await stateFactory(link, token.rawToken);
-			return fetch(state);
+			return fetch(state, true);
```
tenbradley, line 109: I'm a bit concerned about bypassing the cache here for every instance and not just for a particular object or instance.

state/observable/SirenSubEntities.js
```diff
@@ -54,8 +54,7 @@ export class SirenSubEntities extends Observable {
 	}
 
 	async setSirenEntity(sirenParsedEntity) {
-		const subEntities = sirenParsedEntity && sirenParsedEntity.getSubEntitiesByRel(this._rel);
-		if (!subEntities || subEntities.length === 0) return;
+		const subEntities = (sirenParsedEntity && sirenParsedEntity.getSubEntitiesByRel(this._rel)) || [];
```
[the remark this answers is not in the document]
stevenao: there is a problem that as a result of deletion of last sub-entity. The sub-entities observer property is not being updated. It has to do with line 84 and 85 not being called.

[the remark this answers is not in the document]
tenbradley: Okay. That makes sense to me.

state/HypermediaState.js
stevenao: I am seeing a cache primer link is being called with browser cache instead of fetching from the server. We need to refresh the collection after items are deleted.

zinabat: Since this is sitting in the `handleCachePriming` call only, I'm okay with the assumption that we want to refetch all links that we pass to this call.

apalaniuk: It seems a bit unfortunate if cache priming won't respect existing caches - if we're going the "push things to clients" route, I want to say that services should [set`Location` headers on responses to requests using unsafe methods indicating invalidated resources](https://datatracker.ietf.org/doc/html/rfc7234#section-4.4), but I suppose that won't work for invalidation across different domains (hostnames).

apalaniuk: It seems a bit unfortunate if cache priming won't respect existing caches, as there's a potential overpush problem - if we're going the "push things to clients" route, I want to say that services should setLocation headers on responses to requests using unsafe methods indicating invalidated resources, but I suppose that won't work for invalidation across different domains (hostnames).

apalaniuk (approved):

PR #109, opened 2021-07-26
fix: updating state push to push actions and routes seperately

tenbradley:
## Context:
[US130231](https://rally1.rallydev.com/#/357252966636d/custom/367300408400?detail=%2Fuserstory%2F605033238381&fdp=true?fdp=true)
Attempting to fix how the actions are being pushed.

This is also a minor update to the code to separate the routes and actions to be pushed separately to allow us to check that the actions are ready before pushing.

## Quality:
Tested with learning paths. Does not break if the stopUpdates variables are removed.
Tested with work to do.


tenbradley:
This update does not break the components we have checked.

state/HypermediaState.js
```diff
@@ -127,17 +127,22 @@ class HypermediaState extends Fetchable(Object) {
...
+			await state.push();
+		}
 
-		this._routedStates().forEach(action => actions.push(action));
 		const actionMap = this._getMap(this._decodedEntity, observableTypes.action);
 
 		actionMap.forEach(action => actions.push(action));
 		for await (const action of actions) {
-			await action.push();
+			if (action._readyToSend) {
```
AKobets, line 142: what is _readyToSend and could it introduce flakiness?

state/HypermediaState.js
```diff
@@ -127,17 +127,22 @@ class HypermediaState extends Fetchable(Object) {
 	}
 
 	async push() {
-		this._stopUpdates = true;
 		const actions = [];
+		const routedStates = [];
+
+		this._routedStates().forEach(state => routedStates.push(state));
+		for await (const state of routedStates) {
```
AKobets, line 134: could perfromance here be improved with await promise.all instead of awaiting for each promise sequentually ?

state/HypermediaState.js
apalaniuk: [This already exists in `push`](https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/776f844a04834085c3e94646df6d875d0c83801f/state/observable/SirenAction.js#L68), so I imagine this is to only await if there's asynchronicity. If so, it's valid to await non-async calls, and `_readyToSend` implies it's a private/protected property to me as well.

apalaniuk: [This already exists in `push`](https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/776f844a04834085c3e94646df6d875d0c83801f/state/observable/SirenAction.js#L68), so I imagine this is to only await if there's asynchronicity. If so, it's valid to await non-async calls, and `_readyToSend` implies it's a private/protected property to me as well, so the check could be removed if so.

(Related, I suspect that `actions` is not an async iterable, but that predates this change, for sure)

state/HypermediaState.js
```diff
@@ -127,17 +127,22 @@ class HypermediaState extends Fetchable(Object) {
 	}
 
 	async push() {
-		this._stopUpdates = true;
 		const actions = [];
+		const routedStates = [];
+
+		this._routedStates().forEach(state => routedStates.push(state));
```
apalaniuk, line 133: nit, and not as relevant if you go with `Promise.all()`: you could use `this._routedStates()` directly in the `for...of` as it's synchronous, will only be evaluated once and isn't used elsewhere, then remove the `await` in the `for await` as it's not an async iterable (I see this pattern is used elsewhere, so reasonable to have consistency, but the original pattern is also on a non-async iterable, so it implies things that aren't quite true, I think). 

If you want to do a shallow copy of an array, you can also use destructuring, eg.
```javascript
const routedStates = [...this._routedStates()];
```
Which is not only terser, but typically much faster than a `forEach`.

BhamelinD2L:
Checked Discover Rules after our meeting this morning - No issues introduced for us. Thanks!

stevenao:
Actually, it is getting into an infinite looping when trying it in Face Quiz delete question. I am suspecting there is some circlar route references there. 

You can reproduce it by going into Face Quiz and then delete a question. After hitting the delete button at the confirmation dialog, the UI will freeze.

tenbradley:
@stevenao I've just added in a map to track which states and actions have already been pushed. This might resolve your issue.

stevenao:
> @stevenao I've just added in a map to track which states and actions have already been pushed. This might resolve your issue.

Yes, the issue is not showing in my local test anymore. Thanks.

state/HypermediaState.js
zinabat: `_readyToSend` is a property that tells us whether the action has been prepared for sending to the server. Many of the actions we observe we in fact never commit - this property should probably be converted to a public property since we are accessing it from a parent state.

We have to wait for each call to the server to finish because not doing so introduces race conditions

state/HypermediaState.js
zinabat: Currently we have to wait for each call to finish so as not to introduce race conditions, but I'm wondering now if there is a better way to do this

state/HypermediaState.js
zinabat: ah, I just realized you are right, and we are actually not using an async iterable..

I thought the race conditions were a thing, but maybe they just aren't anymore?

apalaniuk: That await is only on the iterable (basically `const state = await routedStates`), so won't have any effect on the flow of the iteration in this case. It's also not invalid, just makes things more confusing than need be if the iterable will never be async.

state/HypermediaState.js
apalaniuk: Can be dropped from here since [`SirenAction::push` still has the check](https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/1664ae5f66f1afb3dae04ed93256a37b0d89f36b/state/observable/SirenAction.js#L72), I think?

state/TrackingMap.js
```diff
@@ -0,0 +1,25 @@
+
+export class TrackingMap {
+
+	constructor() {
+		this._seen = new Map();
+		this._remaining = new Set();
+	}
+
+	checkAndTrack(entity) {
```
apalaniuk, line 9: This tracks actions rather than entities, right? (Though this is generic enough that maybe this is "entity" in the generic, non-Siren sense)

Not too familiar with this code, but are object references sufficient? Is it possible for cycles to be caused due to different object references but the same URL+ action?

(nit: I might return _true_ on success (item tracked) and false otherwise.)

tenbradley:
There's a test that is failing now.
- The test is failing regardless of if the tracking is turned on or off.
- The test that is failing hasn't been changed.

Thus, I don't think the test is failing due to this pr. I want to skip it and add it to the list of tests to be figured out for this story: 
[US130579](https://rally1.rallydev.com/#/357252966636d/workviews?detail=%2Fuserstory%2F605767521393&view=b9761155-0e69-4bb7-a27a-cee9c978c119&fdp=true?fdp=true)

test/state/HypermediaState.test.js
```diff
@@ -435,6 +435,34 @@ describe('HypermediaState class', () => {
 			});
 
 		});
+
```
tenbradley, line 438: I need to make this test better. It currently times out if it fails, but should be better. I have been struggling with getting it to do what I want.

state/HypermediaState.js
tenbradley: It probably should be. We thought this putting here would help with an error that was occurring, but I think it makes as much sense to just be in the action.

state/TrackingMap.js
tenbradley: The reason this tracking is being added is that it is possible for there to be a loop in entities which caused an infinite loop.

The things currently being stored/tracked are HypermediaStates and Actions. This is being used to only for push these entities at the moment. I think that object references are sufficient but could be convinced otherwise.

I'm conflicted about this function. I like there being one function to check if it's tracked and add it if it's not, but these are also different actions which possibly should be split into two functions.

apalaniuk: I think it's a reasonable approach - it's basically the try pattern, with the only failure case being "it already exists". My only concern (out of some ignorance) is the suitable definition of identity to prevent cycles, which if they're caused by the same object reference being used for a push, this makes sense. :+1:

test/state/HypermediaState.test.js
```diff
@@ -435,6 +435,34 @@ describe('HypermediaState class', () => {
...
+			}]
+		};
+
+		const state = await stateFactory(entityHref, 'token');
+		const observer = {};
+
+		state.addObservables(observer, observable);
+		await state.setSirenEntity(SirenParse(JSON.stringify(childEntity)));
+
+		await state.push();
```
zinabat, line 465: instead of awaiting it and using that as a test pass, you'll want to test for at least two things:
1. That the call to `state.push` resolves at all. You can do this a few ways, but the one I tend to prefer is using the extension `chai-as-promised`: https://www.chaijs.com/plugins/chai-as-promised/
2. That the call to the first entity's `push` happens exactly once, or that both happen exactly once. You'll need to use `sinon.spy` to wrap your push calls in order to do this.

zinabat: Secondarily, I don't think that the `state.push` call will do anything in the first place because no actions have been committed.

test/state/HypermediaState.test.js
```diff
@@ -435,6 +435,34 @@ describe('HypermediaState class', () => {
 			});
 
 		});
+
+	});
+
+	it('entity is a subentity of itself', async() => {
```
zinabat, line 441: Remember to title tests as what the behaviour is. In this case,
```suggestion
	it('pushes each state only once when there is a circular dependency', async() => {
```

test/state/HypermediaState.test.js
tenbradley: I don't care that there are no actions as I'm trying to test what happens when there is a circular dependency between entities, in this case, with itself.

I was trying to setup a spy based on the other test for pushes, but was struggling with getting it to do what I wanted.  In this case, push will be called twice. The second time it will see that it's already been pushed and thus not run the body of the push function. I'll poke this some more and try to figure it out.

test/state/HypermediaState.test.js
```diff
@@ -435,6 +435,40 @@ describe('HypermediaState class', () => {
...
+
+		state.addObservables(observer, observable);
+		await state.setSirenEntity(SirenParse(JSON.stringify(childEntity)));
+
+		const pushSpy = sinon.spy(state, 'push');
+
+		console.warn('if a timeout occurs here, it is due to an infinite loop from pushing to a circular dependency. Test in HypermediaState.test.js.');
+		await state.push();
+
+		assert(pushSpy.calledTwice);
```
zinabat, line 471: Can you put a note here after the assert so that we know why it's called twice? e.g., 
```suggestion
		assert(pushSpy.calledTwice, "Spy is called twice, where second call does nothing");
```

zinabat (approved):

PR #110, opened 2021-07-28
feat: Add error handling for the HypermediaStateMixin when a fetch error occurs

alogan519:


state/Fetchable.js
```diff
@@ -6,6 +6,14 @@ import { refreshToken } from './token.js';
  */
 export class FetchError extends Error {}
 
+export class FetchServerError extends FetchError {
+	constructor(status, message) {
+		super(status);
```
apalaniuk, line 11:
```suggestion
		super(message);
```

Since `FetchError` is basically [a marker interface for `Error`](https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/776f844a04834085c3e94646df6d875d0c83801f/state/Fetchable.js#L7), which [has `message` as its first parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error). Could then drop `errorMessage` in favour of the standard `message` property.

state/observable/SirenAction.js
```diff
@@ -50,7 +50,7 @@ export class SirenAction extends Fetchable(Observable) {
 
 	async onServerResponse(json, error) {
 		if (error) {
-			throw new FetchError(error);
```
apalaniuk, line 53: The fact that this is a marker interface/class makes me a bit concerned that something is doing `instanceof` checks (I'm not sure for any other reason for this), but I can't find anything in any of our orgs that does this... so seems safe? More sensical error logging your way.

state/fetch.js
```diff
@@ -63,7 +64,13 @@ async function performServerFetch(fetchable, bypassCache) {
 
 	const response = await fetch.fetch(fetchable.href, { headers, body: fetchable.body, method: fetchable.method });
 	if (!response.ok) {
-		throw response.status;
+		let json;
+		try {
+			json = await response.json();
+		} catch (e) {
+			throw new FetchServerError(response.status, '');
```
apalaniuk, line 71: I might consider capturing the inner error, or its message at least - I'm sure failures to get a JSON body will be logged in the console automatically, so wouldn't be a problem for humans doing manual testing, but it could make traceability easier in some contexts (Elastic RUM down the road?). But, all infinitely better than throwing the non-`Error` as done before though.

state/Fetchable.js
alogan519: Thanks, sounds good.

state/fetch.js
alogan519: Ya my thought here was that somnething like a 404 response we don't get a response.json so I was worried we'd be logging here when it's not that unusual.

state/Fetchable.js
alogan519: Oh actually I like that status is the call to super since it's not changing existing behavior and message may be empty.

state/fetch.js
apalaniuk: It would be neat to check the `Content-Type` header to only `.json()` JSON-things and use other methods (if there's any body at all) otherwise, to always capture the actual response. The logging part was speculative so, not really tied to decisions around that too much (not a blocker).

I also wonder if any clients were using the response status directly. It seems it would always be wrapped in a `FetchError`, and using the message contents would be fragile anyway so shouldn't be done, just a thought. I can't find any, so 🤷

state/Fetchable.js
apalaniuk: Yeah, not changing existing behaviour might just be the best thing (even if I think the original way is odd).

state/observable/SirenAction.js
apalaniuk: I guess this will always be either a `FetchServerError` or `FetchError` anyway, so would still be safe if so.

apalaniuk:
The [comment on response deserialization](https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/110#discussion_r679272737) is the only one I feel strongly about, but not going to block if others who are more literate in this area think it's reasonable.

state/Fetchable.js
apalaniuk: Yeah, not changing existing behaviour might just be the best thing.

apalaniuk (approved):

PR #114, opened 2021-08-16
fix: bypassing cache for summon actions

tenbradley:


state/observable/SirenSummonAction.js
```diff
@@ -55,6 +55,7 @@ export class SirenSummonAction extends Routable(SirenAction) {
 		this._routes.forEach((route, observer) => {
 			this.routedState.addObservables(observer, route);
 		});
+		this.routedState.byPassCache();
```
BhamelinD2L, line 58: This line is done optionally on line 90 as an argument by specifying bypassCache=true to the summon call. By setting it true there, it clears it before it gets here. If Polaris and everyone else is fine with it always being true, I am as well, but if that's the case, do we need to remove that argument from summon()?

PR #116, opened 2021-09-17
docs: Add JSdoc comments for two commonly imported functions

zinabat:
## Context
Added JSdoc comments for easier usage of the engine with standard IDEs.

When hovering over a function that is documented with JSdoc, a developer can see the information at a glance. They can also be used to generate documentation in a README, but that will come later.

### Result
Screen Shot 2021-09-17 at 11 43 32 AM

Screen Shot 2021-09-17 at 11 43 41 AM


bearfriend (approved):
🍇

PR #118, opened 2021-10-05
feat: US131709 - Include the message for FetchServerErrors

maggiesazio:
https://rally1.rallydev.com/#/?detail=/userstory/609149953873&fdp=true


state/fetch.js
```diff
@@ -68,6 +68,10 @@ async function performServerFetch(fetchable, bypassCache) {
 			const message = await response.text();
 			throw new FetchServerError(response.status, message);
 		} catch (e) {
+			if (e instanceof FetchServerError) {
```
maggiesazio, line 71: Trying to use the error message in the LTI component but it was getting stripped out here

alogan519: ugh ya that was probably my bad.

Maybe like 
`if (!response.ok) {
  let message = ''
  try {
    message = await response.text();
  } catch(e) {
    throw new FetchServerError(respoinse.status, '');
  }
  throw new FetchServerError(response.status, message);
}
`

or

`if (!response.ok) {
  let message = ''
  try {
    message = await response.text();
  } finally {
    throw new FetchServerError(respoinse.status, message);
  }
}
`

PR #125, opened 2021-12-15
DO NOT MERGE: BREAKING CHANGE: upgrade to lit-html 2

annabelsegalld2l:
New TemplateResult is a type: https://lit.dev/docs/api/templates/#TemplateResult with fields {_$litType$: T, strings: TemplateStringsArray, values: Array<unknown>}

Old TemplateResult is a class: https://lit.dev/docs/v1/api/lit-html/templates/#TemplateResult with constructor parameters (strings, values, type, processor)

Essentially, I've reconstructed the TemplateResult type into the HypermediaResult class. It seems like lit 2 just didn't add a class with the litType so I needed to manually keep the field in for the tests to pass.


framework/lit/hypermedia-components.js
```diff
@@ -22,5 +21,5 @@ export function customHypermediaElement(tag, elementClass, pseudoTag, hypermedia
  * @returns {HypermediaResult} The template result ready for rendering
  */
 export function html(strings, ...values) {
```
annabelsegalld2l, line 23: Didn't change the API parameters so consumers should not be affected by this change.

framework/lit/HypermediaResult.js
```diff
@@ -15,10 +15,12 @@ export function custom(tag, elementClass, pseudoTag, hypermediaClasses, options)
  * Extension to the TemplateResult class from lit-html
  * Contains the logic for replacing pseudotags with tags
  */
-export class HypermediaResult extends TemplateResult {
+export class HypermediaResult {
```
annabelsegalld2l, line 18: It really seems like this should have been more prominent in the docs.....
TemplateResult was changed from a class to a type, so you can't extend it anymore (many hours wasted 🤦).
What I've done here is keep the Hypermedia class, and manually assign the string, litType and values fields.

package.json
```diff
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "@babel/eslint-parser": "^7.14.5",
-    "@open-wc/testing": "^2.5.29",
+    "@open-wc/testing": "^3",
```
annabelsegalld2l, line 22: Bumped to the next major version while we are in here.

framework/lit/HypermediaResult.js
dlockhart: Crazy, sorry you had to go through that! Does this mean then that if we removed the lit-html-/lit-element upgrade bits from this PR, it could merge now and not be a breaking change?

i.e. could we merge part of this today and release it and then upgrade Lit later when we do everything else?

annabelsegalld2l: Ah, interesting idea. It _should_ work... the tests were actually pretty thorough and caught some errors. I can do some comparisons with/without the change on some pages to double-check it doesn't break anything.

annabelsegalld2l:
Closing in favour of a non-breaking PR for now. Once it is merged, I will open another PR with no merge conflicts.

PR #126, opened 2021-12-15
US134612 - remove TemplateResult import from HypermediaResult in preparation for Lit 2

annabelsegalld2l:
This is an alternate approach to my previous PR where I tried to emulate the old TemplateResult class structure. Instead, we can manually create an html object with the built-in properties already added. This prevents us from having to deal with the internals like the templateProcessor or _litType, making the code cleaner.

Context: TemplateResult in Lit 2 is a type: https://lit.dev/docs/api/templates/#TemplateResult with fields {_$litType$: T, strings: TemplateStringsArray, values: Array}

Old TemplateResult is a class: https://lit.dev/docs/v1/api/lit-html/templates/#TemplateResult with constructor parameters (strings, values, type, processor)


dlockhart (approved):
Nice! Awesome that we can get this in right away before the upgrade. 💯

PR #128, opened 2022-01-04
Revert "US134612 - remove TemplateResult import from HypermediaResult in preparation for Lit 2"

annabelsegalld2l:
Reverts BrightspaceHypermediaComponents/foundation-engine#126.

Margaret from Quizzing reported some issues with tokens being appended to the href instead of the token parameter. I am reverting this change while I work on a fix to unblock any consumers.


annabelsegalld2l:
I believe I've found the underlying issue! I'll close this PR and make a new one with just the fix.

PR #131, opened 2022-01-19
Lit 2 prep - invalid template array errors - Force the html objects to have strings.raw property

annabelsegalld2l:
Lit 2 adds a security check to its render function to ensure the html 'strings' property is the [TemplateStringsArray](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.templatestringsarray.html) type. We are currently getting this [error](https://github.com/lit/lit/blob/main/packages/lit-html/src/lit-html.ts#L661). Although the class gets a TemplateStringsArray passed in, we are required to modify it, creating a `string[]` instead. As we are not using Typescript, we currently have no way to cast from the string array to the correct type.

One workaround we can do is add a "raw" property to our strings element, which will pass the security check for us. Note: **this is a pretty hacky solution. The recommended long-term solution would be to use TS to ensure type-safety. However, this does not make sense to do now until we are fully on Lit 2. ** The types and class definitions change drastically, so we would have to do rewrites in the migration. This essentially acts as a stopgap to allow us to move forward in development for repos blocked on this fix.


dlockhart (approved):
Agreed this isn't ideal, but if it can get us through the upgrade it seems like the only option. Post-upgrade, we could convert even just the required files to TS and then compile it to JS before publishing to NPM -- so other downstream projects don't need to worry about TS.

PR #137, opened 2022-02-09
Lit 2 fix: Skeletons need to manually add the 'raw' property

annabelsegalld2l:
I had previously added this function to prevent `invalid template array` errors in Lit 2 rendering. I was testing foundation-components and found the issue also happens with HM skeletons, since they are passing a stringArray rather than the TemplateStringArray type. 

I also added some tests that would fail in lit 2 without the skeleton change, as they would not be able to render.


test/framework/lit/hypermedia-result.test.js
```diff
@@ -1,11 +1,12 @@
...
+import { expect, fixture }  from '@open-wc/testing';
+import { LitElement, html as lithtml } from 'lit-element';
 import { default as fetchMock } from 'fetch-mock/esm/client.js';
-import { html as lithtml } from 'lit-element';
+import { HypermediaStateMixin } from '../../../framework/lit/HypermediaStateMixin.js';
 
 class TestComponent { }
-class TestHMComponent { }
-class TestHMComponentExtend { }
+class TestHMComponent extends HypermediaStateMixin(LitElement) { }
```
annabelsegalld2l, line 8: These need to extend the HypermediaStateMixin to actually render. Previously, we were not testing rendering at all, just creating the HTML object. These tests will help us catch any render errors in the future as well.

framework/lit/HypermediaResult.js
```diff
@@ -226,6 +226,6 @@ export class HypermediaResult {
 
 	_skeletonRender(components) {
 		const tag = components.skeletonTag();
-		return tag ? html([`<${tag} skeleton></${tag}>`]) : null;
+		return tag ? html(this._getTemplateStringArray([`<${tag} skeleton></${tag}>`])) : null;
```
annabelsegalld2l, line 229: This helper adds the `raw` property to the string array that Lit's render function is looking for.

tenbradley (approved):
I'm happy with these changes.

PR #144, opened 2022-03-01
fix: make state when re-connected

mdulat:
This one was a doozy. When a quiz question was moved within the d2l-lsit (either by drag & drop or MoveTo action), the `HypermediaState`'s `_observers` property became empty. As a result, future updates of the item did not get updated in the UI because there were no longer any observers attached to the item component.

The cause was that when the item is moved, the component is disconnected, then reconnected quickly afterwards. When reconnecting, the `_state` property was still there, so the state was not recreated. But shortly after that the `_state` was disposed.

The fix is to make sure that if we are in the process of disposing while connecting the component, the state is re-created.

I need some hmc experts to please check that this isn't going to break something else!

To reproduce the issue, you can check a course on Quad that has a quiz with questions, e.g. in course "FACE opt-in (Competencies)", quiz "Quiz with questions":

1. launch face quiz w/ questions
2. select a question
3. toggle bonus (see update correctly)
4. w/ question still selected, move to top or bottom
5. toggle bonus (do not see update)


mdulat:
Probably need review by code owners for this one please! @tenbradley @AKobets

AKobets:
@mdulat what testing have you performed to ensure it works and there is no regression ?

tenbradley:
I've manually tested learning paths, work to do, and discover rules with this change. I see no problems introduced based on this pr.

mdulat:
> Would it make sense to store `this._state` in a temporary variable, set `this._state` to null, and then call dispose with the temporary? This might be an excessively cautious approach.

I'm all for excessive caution :). This makes sense, and it works fine for me locally. I'll update the PR.

tenbradley (approved):

PR #146, opened 2022-03-08
fix: avoid setting state to null

mdulat:
Turns out setting state to null breaks Work To Do. I think it's because when making the state, after the fetch call we have [an async function](https://github.com/BrightspaceHypermediaComponents/foundation-engine/blob/master/framework/lit/HypermediaStateMixin.js#L97), and sometimes that state is set to null after fetching but before the async function is called. So reverting this back to the original fix, which I think is what you tested @tenbradley with learning paths, work to do, and discover rules, right? It is working for me locally (tested W2D and original quizzing bug). 

([original fix](https://github.com/BrightspaceHypermediaComponents/foundation-engine/pull/144))


dcprime (approved):

tenbradley (approved):

tenbradley:
Sounds good. My over cautiousness caused an issue. 😞

